PR #2, opened 2020-09-11
Cache validator public keys for given era

sgladkov:
Optimize ValidatorManager::GetValidatorsPublicKeys. This function is called from GetValidatorForEra and GetValudatorIndex and takes a lot of time. The values it returns are not changed with time,  so we can store them in memory to have quick access.

Please see to the Dictionary object declaration,  Is it valid to make it static in this situation?

### DIFF
--- a/src/Lachain.Core/Blockchain/Validators/ValidatorManager.cs
+++ b/src/Lachain.Core/Blockchain/Validators/ValidatorManager.cs
@@ -13,7 +13,7 @@ namespace Lachain.Core.Blockchain.Validators
     {
         private readonly ISnapshotIndexRepository _snapshotIndexRepository;
         private static readonly ILogger<ValidatorManager> Logger = LoggerFactory.GetLoggerForClass<ValidatorManager>();
-        private static Dictionary<long, IReadOnlyCollection<ECDSAPublicKey>> _pubkeyCache = new Dictionary<long, IReadOnlyCollection<ECDSAPublicKey>>();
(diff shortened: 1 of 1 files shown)


src/Lachain.Core/Blockchain/Validators/ValidatorManager.cs
```diff
@@ -13,6 +13,7 @@ public class ValidatorManager : IValidatorManager
     {
         private readonly ISnapshotIndexRepository _snapshotIndexRepository;
         private static readonly ILogger<ValidatorManager> Logger = LoggerFactory.GetLoggerForClass<ValidatorManager>();
+        private static Dictionary<long, IReadOnlyCollection<ECDSAPublicKey>> _pubkeyCache = new Dictionary<long, IReadOnlyCollection<ECDSAPublicKey>>();
```
sgladkov, line 16: Is it valid to make it static in this situation?

Skird: It's technically valid provided this class is used as singleton (it currently is). However cleaner solution IMO would be
`private readonly Dictionary<long, IReadOnlyCollection<ECDSAPublicKey>> _pubkeyCache`
This way one can create multiple instances of class (e.g. may be useful in tests) which won't interfere

sgladkov: Fixed

sgladkov:
@Skird , fixed,  re-review, please

PR #12, opened 2021-01-26
FastSync

dushyantb:


docker-compose.yml
```diff
@@ -5,69 +5,74 @@ services:
       network: host
       context: ./
       args:
-        CONFIG: src/Lachain.Console/config0.json
-        WALLET: src/Lachain.Console/wallet0.json
+        CONFIG: src/Lachain.Console/poc/config01.json
```
sgladkov, line 8: There is no such file in git. Do we need this docker files changes?

docker-compose.yml
```diff
@@ -5,69 +5,74 @@ services:
...
-#    - "7071:7070"
+    # ports:
+    #   - "7070:7070"
+    #   - "5050:5050"
+    # ports:
+    #   - "5050:5050"
+    #   - "7071:7070"
     network_mode: "host"
-    restart: always
+    restart: "no"
```
sgladkov, line 23: I don't sure,  but looks like this property is used by cloud nodes too (for example,  to restart them after keys rescan)

docker-compose.yml
```diff
@@ -5,69 +5,74 @@ services:
...
 
   node1:
     build:
       network: host
       context: ./
       args:
-        CONFIG: src/Lachain.Console/config1.json
-        WALLET: src/Lachain.Console/wallet1.json
+        CONFIG: src/Lachain.Console/poc/config02.json
```
sgladkov, line 30: There are no this files in git

src/Lachain.Console/Application.cs
```diff
@@ -113,6 +134,89 @@ public void Start(RunOptions options)
                 Thread.Sleep(1000);
         }
 
+        private Block BuildNextBlock(IStateManager _stateManager, IPrivateWallet _wallet,
```
sgladkov, line 137: I don't sure,  but do we need this methods here? I can't find where are they called.)

sgladkov (commented):
Looks like working for me,  but there are several places where the "debug" code is added,  I think it is better to remove this code before merging.  But we can merge it now and deploy it in testnet and remove this code later if it is more convenient

docker-compose.yml
dushyantb: added the POC folder in the repo... we need changes such that every peer in the network is having unique RPC port

docker-compose.yml
```diff
@@ -5,69 +5,74 @@ services:
...
-        CONFIG: src/Lachain.Console/config0.json
-        WALLET: src/Lachain.Console/wallet0.json
+        CONFIG: src/Lachain.Console/poc/config01.json
+        WALLET: src/Lachain.Console/poc/wallet01.json
+        PRVH: src/Lachain.Console/poc/prv_h01.txt
       dockerfile: Dockerfile
     stdin_open: true
     environment:
-      - LOG_LEVEL=Trace
+      - LOG_LEVEL=Debug
```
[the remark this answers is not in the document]
dushyantb: Reverted back to `Trace`

docker-compose.yml
dushyantb: this is for in case if the container gets closed by any error, it will restart automatically...
Reverted back to `always`

docker-compose.yml
dushyantb: added

src/Lachain.Console/Application.cs
dushyantb: it was redundant code... removed

sgladkov (approved):
LGTM now.)

src/Lachain.Core/RPC/HTTP/FastSyncService.cs
```diff
@@ -0,0 +1,536 @@
...
+                    trieHashMap.GetSerializedNodes(version).ToList());
+
+                _repoBlocks.Add(repoType, t);
+            }
+
+            Logger.LogTrace($"End: getSerializedNodesAndIds");
+        }
+
+        [JsonRpcMethod("getBlocks")]
+        private JObject? GetBlocks(ulong repoType, int offset)
```
sgladkov, line 110: Looks like this implementation can cause issues if several clients call it simultaneously

src/Lachain.Console/Application.cs
```diff
@@ -92,15 +99,33 @@ public void Start(RunOptions options)
...
+            Thread.Sleep(10000);
+
+            if (blockManager.GetHeight() == 0)
+            {
+                blockSynchronizer.StartFastSync();    
+            }
+            
+            // consensusManager.Start((long)blockManager.GetHeight() + 1);
+            // validatorStatusManager.Start(false);
+
```
sgladkov, line 114:
```suggestion
            if (blockManager.GetHeight() == 0)
            {
                blockSynchronizer.StartFastSync();    
            }
            
            consensusManager.Start((long)blockManager.GetHeight() + 1);
            validatorStatusManager.Start(false);
```
This code is working for me on local testnet (start 4 nodes-validators, stop 1 of them,  clear its database,  start it again after 50 blocks are published)

src/Lachain.Core/RPC/HTTP/FastSyncService.cs
dushyantb: there would be only 1 client call at the time and won't be any simultaneous calls. this has been taken care by the `handShake` mehod

src/Lachain.Console/Application.cs
sgladkov, line 114:
```suggestion
            if (blockManager.GetHeight() == 0)
            {
                blockSynchronizer.StartFastSync();    
            }
            
            consensusManager.Start((long)blockManager.GetHeight() + 1);
            validatorStatusManager.Start(false);
```
This variant is working for me in local testnet,  in other case consensusManager causes errors during the sync

dushyantb: This will not be the final state, this I've used so that on local I can test the FastSync

PR #17, opened 2021-04-15
@LACHAIN-332 : Fix keygen tx generation issues

sgladkov:


src/Lachain.Core/Vault/KeyGenManager.cs
```diff
@@ -64,10 +64,15 @@ private void BlockManagerOnSystemContractInvoked(object _, InvocationContext con
         {
             if (context.Receipt is null) return;
             var highestBlock = _blockSynchronizer.GetHighestBlock();
-            var willParticipate =
+            var willParticipate = 
                 !highestBlock.HasValue ||
-                GovernanceContract.IsKeygenBlock(highestBlock.Value) &&
+                GovernanceContract.IsKeygenBlock(context.Receipt.Block) &&
```
sgladkov, line 69: Take a look at this change: I don't sure I understand this condition good enough, but I think it is more correct than before (and it is really works for small cycles,  I've checked it with debugging exactly this place)

Skird (approved):

PR #21, opened 2021-06-07
@LACHAIN-362: Implemented the function

dushyantb:
https://latoken.atlassian.net/browse/LACHAIN-362

Function implemented and @soloway2010 has verified the same and working fine.

### DIFF
--- a/src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
+++ b/src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
@@ -308,6 +308,14 @@ namespace Lachain.Core.Blockchain.VM
             };
             frame.InvocationContext.Snapshot.Events.AddEvent(ev);
         }
+        
+        public static void Handler_Env_GetAddress(int resultOffset)
+        {
+            var frame = VirtualMachine.ExecutionFrames.Peek() as WasmExecutionFrame
+                        ?? throw new InvalidOperationException("Cannot call KECCAK256 outside wasm frame");
+            var result = (frame.CurrentAddress).ToBytes();
+            SafeCopyToMemory(frame.Memory, result, resultOffset);
+        }
 
         private static FunctionImport CreateImport(string methodName)
         {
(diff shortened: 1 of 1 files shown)


src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
```diff
@@ -308,6 +308,14 @@ public static void Handle_Env_WriteEvent(int signatureOffset, int valueOffset, i
             };
             frame.InvocationContext.Snapshot.Events.AddEvent(ev);
         }
+        
+        public static void Handler_Env_GetAddress(int resultOffset)
+        {
+            var frame = VirtualMachine.ExecutionFrames.Peek() as WasmExecutionFrame
+                        ?? throw new InvalidOperationException("Cannot call KECCAK256 outside wasm frame");
```
soloway2010, line 315: Perhaps "?? throw new InvalidOperationException("Cannot call GetAddress outside wasm frame");" is better?

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
```diff
@@ -308,6 +308,14 @@ public static void Handle_Env_WriteEvent(int signatureOffset, int valueOffset, i
...
+            var result = (frame.CurrentAddress).ToBytes();
+            SafeCopyToMemory(frame.Memory, result, resultOffset);
```
soloway2010, line 317: Add returned value check?
```
if (!ret)
    throw new InvalidContractException("Bad call to (get_address)");
```

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
dushyantb: yes... will update the code.

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
dushyantb: No, as we are not returning any values here, and execution frame will always be having the `currentAddress`. So, essentially this check is not required .

soloway2010 (approved):

Skird (approved):

PR #24, opened 2021-06-22
made some changes to run the tests successfully in windows machine

Arghya115:
Lachain.CoreTest and Lachain.StorageTest were failing, as _container.Dispose() was not placed properly in some files
. So when TestUtils.DeleteTestChainData() was called, a file(LOCK) was still open in the process and errors occurred. Tweaked positions of these two methods in some files. Now Lachain.StorageTest runs properly but Lachain.CoreTest sometime runs properly and sometime 4/5 tests crash among 16 tests.

### DIFF
--- a/test/Lachain.UtilityTest/TestUtils.cs
+++ b/test/Lachain.UtilityTest/TestUtils.cs
@@ -9,7 +9,6 @@ using Lachain.Proto;
 using Lachain.Utility;
 using Lachain.Utility.Utils;
 using Nethereum.Util;
-using System.Diagnostics;
 
 
 namespace Lachain.UtilityTest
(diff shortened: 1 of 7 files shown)


test/Lachain.UtilityTest/TestUtils.cs
```diff
@@ -9,6 +9,8 @@
 using Lachain.Utility;
 using Lachain.Utility.Utils;
 using Nethereum.Util;
+using System.Diagnostics;
```
sgladkov, line 12: It is better to revert this file to original code if we don't make any changes in it really

test/Lachain.UtilityTest/TestUtils.cs
```diff
@@ -37,7 +39,7 @@ public static TransactionReceipt GetRandomTransaction()
         public static void DeleteTestChainData()
         {
             var chainTest = Path.Join(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "ChainTest");
-            if (Directory.Exists(chainTest)) Directory.Delete(chainTest, true);
+            if (Directory.Exists(chainTest))  Directory.Delete(chainTest, true);
```
sgladkov, line 42: Revert this line too

sgladkov (changes requested):

sgladkov (approved):

PR #26, opened 2021-06-25
369 feature get msg value

dushyantb:


src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
```diff
@@ -116,6 +121,8 @@ public static void Handler_Env_WriteLog(int offset, int length)
                 throw new InvalidContractException("Bad call to call function");
             var address = addressBuffer.Take(20).ToArray().ToUInt160();
             var value = SafeCopyFromMemory(frame.Memory, valueOffset, 32)?.ToUInt256().ToMoney();
+            var msgValue = SafeCopyFromMemory(frame.Memory, valueOffset, 32)?.ToUInt256();
```
soloway2010, line 124: There is no need in `msgValue` variable as it copies the value of `value` and it calls `SafeCopyFromMemory` the second time with the same args

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
```diff
@@ -316,6 +323,17 @@ public static void Handler_Env_GetAddress(int resultOffset)
             var result = (frame.CurrentAddress).ToBytes();
             SafeCopyToMemory(frame.Memory, result, resultOffset);
         }
+        
+        public static void Handler_Env_GetMsgValue(int dataOffset)
+        {
+            Logger.LogInformation($"Handler_Env_GetMsgValue({dataOffset})");
```
soloway2010, line 329: Does it need to be here? I think, it's better to either remove it or add logging to other handlers as well

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
dushyantb: Updated the code in the latest commit. Still we had to keep `value` and `msgValue` both as the type of them are different... `Moeny` and `UInt256` respectively

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
dushyantb: We can have this right now and eventually update the other handlers with logging as well

soloway2010 (approved):

PR #29, opened 2021-07-06
fixed folding logic in case of deletion to remove unnecessary long branch

Arghya115:
In case of deletion, if a node have two child then and one child is deleted and other one is leaf, then we folded this node. But consider it's parent, if that node have only one child, it should be also removed/folded.


HMAshiqulIslam (approved):
LGTM

sgladkov (approved):
For the future it would be great to have some unit tests verifying the changes in storage,  it has a few tests right now, its test coverage is really low.)

PR #35, opened 2021-07-14
added a lru cache and optimized node serialization

HMAshiqulIslam:


Arghya115 (approved):

src/Lachain.Storage/Trie/NodeSerializer.cs
```diff
@@ -12,51 +12,101 @@ public static byte[] ToBytes(IHashTrieNode node)
...
-                        mask,
-                        Enumerable.Range(0, len)
-                            .Select(i => bytes.Slice(1 + 4 + 32 + i * 8, 8).Span.ToUInt64()),
-                        hash
-                    );
+                    var len = (int)BitsUtils.Popcount(mask);
+                    byte[] hash = new byte[32];
+                    for (var i = 0; i < 32; i++)
+                    {
+                        hash[i] = bytes[i + 5];
```
sgladkov, line 79: It is better to check if this bytes are available (check the `bytes` length before the cycle)

src/Lachain.Storage/Trie/NodeSerializer.cs
```diff
@@ -12,51 +12,101 @@ public static byte[] ToBytes(IHashTrieNode node)
...
                 case LeafNode leafNode:
-                    return new byte[] {1}.Concat(leafNode.KeyHash).Concat(leafNode.Value).ToArray();
+                    return new byte[] { 1 }.Concat(leafNode.KeyHash).Concat(leafNode.Value).ToArray();
             }
 
             throw new InvalidOperationException($"Type {node.GetType()} is not supported");
         }
 
-        public static IHashTrieNode FromBytes(ReadOnlyMemory<byte> bytes)
+        public static IHashTrieNode FromBytes(byte[] bytes)
```
sgladkov, line 54: Do we really need to change the data type from read-only type here?

src/Lachain.Storage/Trie/NodeSerializer.cs
```diff
@@ -12,51 +12,101 @@ public static byte[] ToBytes(IHashTrieNode node)
...
+                    ulong[] _children = new ulong[len];
+                    var pos = 32 + 5;
+                    for (var i = 0; i < len; i++)
+                    {
+                        byte sz = bytes[pos];
+                        ulong m = 1;
+                        ulong res = 0;
+                        for (int j = pos + 1; j <= pos + sz; j++)
+                        {
+                            res = res + m * bytes[j];
```
sgladkov, line 90: The same as above: I think it is better to check `bytes` size before and throw a typed exception if smth is wrong

src/Lachain.Storage/Trie/TrieHashMap.cs
```diff
@@ -132,16 +135,23 @@ private IEnumerable<byte[]> TraverseValues(ulong root)
                     break;
                 default:
                     foreach (var child in node.Children)
-                    foreach (var value in TraverseValues(child))
-                        yield return value;
+                        foreach (var value in TraverseValues(child))
```
sgladkov, line 138: It is better to not commit occasional changes like this and several other changes in this file. It makes search in commit history more difficult in the future

sgladkov (changes requested):

PR #38, opened 2021-07-20
added batch tx api

HMAshiqulIslam:


src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
```diff
@@ -189,6 +189,16 @@ private string SendRawTransaction(string rawTx)
             }
         }
 
+        [JsonRpcMethod("fe_sendRawTransactionBatch")]
```
sgladkov, line 192: I think it is better to rename it to `la_sendRawTransactionBatch`, `fe` was for `FrontEnd`,  we don't use ths function in frontend app
```suggestion
        [JsonRpcMethod("la_sendRawTransactionBatch")] 
```

sgladkov (changes requested):

sgladkov (approved):

PR #52, opened 2021-08-06
Lachain 449 string storage

soloway2010:


test/Lachain.CoreTest/IntegrationTests/VirtualMachineTest.cs
```diff
@@ -60,8 +60,8 @@ public void Teardown()
          //   _container.Dispose();
         }
 
-        /*
-        [Test]
+        
+        /*[Test]
```
sgladkov, line 64: Why this test is commented in this pull request? Do we have a test case for storage string? VirtualMachineTests are really helpful at higher levels testing.)

PR #53, opened 2021-08-09
nodehash mismatch bug is fixed, altered logic in SplitNode()

Arghya115:
In SplitNode(), we didn't handle the case if secondFragment<FirstFragment. Fixed that.


sgladkov (approved):
Looks good to me

HMAshiqulIslam (approved):
The logic seems good to me. Really a nice catch to find the bug.

PR #64, opened 2021-08-27
la_validator_info API implementation

dushyantb:


sgladkov (approved):
It is better to not change the lines of code  we are not using in the future, it makes review harder.

PR #81, opened 2021-09-28
Update the Minimal Gas Price

dushyantb:


src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
```diff
@@ -477,8 +477,10 @@ private string Call(JObject opts, string? blockId)
                         Block = snapshot.Blocks.GetTotalBlockHeight(),
                         Transaction = new Transaction{Value = 0.ToUInt256()}
                     });
+                    
+                    var localInvocation = ContractEncoder.Encode("transfer(address,uint256)", source, 0.ToUInt256());
                     var invocationResult =
-                        ContractInvoker.Invoke(destination, systemContractContext, invocation, 100_000_000);
+                        ContractInvoker.Invoke(ContractRegisterer.LatokenContract, systemContractContext, localInvocation, 100_000_000);
```
sgladkov, line 483: Looks like instead of call of system contract here you call your own functionality. I see that you want to call transfer method of the LatokenContract for simple transfer transaction here,  but you should do it for simple transfer exactly,  not by dropping system contract calls. I think you should place it inside the commented block with return at line 445

sgladkov (changes requested):

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
dushyantb: updated the logic in the latest commit

sgladkov (approved):

PR #86, opened 2021-09-30
Verify validator set for peer blocks

sgladkov:


src/Lachain.Core/Blockchain/Operations/BlockManager.cs
```diff
@@ -508,15 +509,31 @@ public OperatingError VerifySignature(BlockHeader blockHeader, Signature signatu
...
 
         [MethodImpl(MethodImplOptions.Synchronized)]
-        public OperatingError VerifySignatures(Block block)
+        public OperatingError VerifySignatures(Block block, bool checkValidatorSet)
         {
             if (!block.Header.Keccak().Equals(block.Hash))
                 return OperatingError.HashMismatched;
             if (_IsGenesisBlock(block))
                 return OperatingError.Ok;
+            if (checkValidatorSet && !VerifyValidatorSet(block.Multisig.Validators, GetHeight()))
```
HMAshiqulIslam, line 518: do you think, instead of GetHeight(), block.index would be better ? Verify signature should not depend on the local state of the chain.

sgladkov: We skip all blocks with height greater than our height + 1 (https://github.com/LATOKEN/lachain/blob/45835999193a70c3ba7a95191bd92505f691085b/src/Lachain.Core/Network/BlockSynchronizer.cs#L141), so we always has GetHeight() = block.Index -1 here

sgladkov: but let's use data from block header,  you are roght

HMAshiqulIslam (approved):
LGTM

PR #91, opened 2021-10-01
Simultaneous unit test

tbssajal:
Added unit tests for api methods in ValidatorServiceWeb3 and BlockchainServiceWeb3.


src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
```diff
@@ -439,42 +456,30 @@ private JArray GetEventsByTransactionHash(string txHash)
...
+            if ((!(fromBlock is null) || !(toBlock is null)) && !(blockhash is null))
                 throw new Exception(
                     "If blockHash is present in in the filter criteria, then neither fromBlock nor toBlock are allowed.");
 
-            var start = (ulong)0;
+            ulong ?start;
+            start = _blockManager.GetHeight();
             if (!(fromBlock is null))
             {
-                if (((string)fromBlock!).StartsWith("0x"))
```
sgladkov, line 457: If we use this functionality,  we should check if we can parse hex strings starting with "0x" and strings that are not started with special prefix. Add this logic to the GetBlockNumberByTag

tbssajal: It is already implemented in Lachain.Utility.Utils.HexUtils.HexToUlong(this string buffer)

src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
```diff
@@ -112,7 +112,13 @@ public class BlockchainServiceWeb3 : JsonRpcService
 
             if(fullTx) 
                 txArray = Web3DataFormatUtils.Web3BlockTransactionArray(txs, block!.Hash, block!.Header.Index);
-
+            else
```
sgladkov, line 115: For the future: do not place different fixes in one pull request,  create other pull request if you need to fix something not related,  pull requests should have a narrow focus,  it really simplifies a work.

sgladkov (changes requested):
Test_Web3Block is failed

src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
tbssajal: Okay, I will keep in mind.

sgladkov (approved):

PR #96, opened 2021-10-13
added a check before incrementing current era

HMAshiqulIslam:


src/Lachain.Core/Consensus/ConsensusManager.cs
```diff
@@ -151,7 +151,19 @@ private void FinishEra()
...
+                    var height = (long) _blockManager.GetHeight();
+                    
+                    if(height >= CurrentEra) 
+                    {
+                        CurrentEra += 1;
+                        Logger.LogTrace($"Current Era is advanced. Current Era: {CurrentEra}");
+                    }
+                    else 
+                    {
+                        Logger.LogWarning($"block height: {height} is smaller than CurrentEra: {CurrentEra}.");
```
sgladkov, line 164: Should we skip EraBroadcaster creation if we skip era advancing?

Ashiq100: I created a new broadcaster out of fear that the finish cycle function may assume a somewhat new broadcaster (for example look at line 209-210 of this changed file). So, I think it's better to have a fresh broadcaster. What do you think ?

PR #101, opened 2021-10-18
Fast sync parallel

Arghya115:
New fault tolerant fast_sync, tested multiple times.


src/Lachain.Core/Network/FastSynchronizerBatch.cs/FastSynchronizerBatch.cs
```diff
@@ -6,24 +6,33 @@
...
                                      VersionFactory versionFactory,
                                      ulong blockNumber)
         {
+            if(Alldone(dbContext))
+            {
+                Console.WriteLine("Fast Sync was done previously\nReturning");
+                return;
+            }
             dbContext.Save(EntryPrefix.NodesDownloadedTillNow.BuildPrefix(), UInt64Utils.ToBytes(0));
             List<string> devnetNodes = new List<string>
```
sgladkov, line 34: Why do we have hardcoded ips here?

sgladkov (changes requested):

src/Lachain.Core/Network/FastSynchronizerBatch.cs/FastSynchronizerBatch.cs
Arghya115: Yes, we can change that and take ips from config file, should I do it now? But in real scenario we need to be able find available peers by some other way.

PR #108, opened 2021-10-29
added function for larc_history

dushyantb:


src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -33,9 +33,22 @@ public void TryAddTransaction(TransactionReceipt receipt)
         {
             if (_watchAddresses.Count(addr =>
                     addr.Equals(receipt.Transaction.To) || addr.Equals(receipt.Transaction.From)) <= 0) return;
-            
-            var data = LoadState();
-            SaveState(data.Concat(receipt.Hash.ToBytes()).ToArray());
+
+            if (receipt.Transaction.Invocation.Length > 0)
```
sgladkov, line 37: We will return at the previous line if tx contains our address in Invocation only,  so this condition will not be checked in this case

sgladkov (changes requested):
Change condition check order or remove return in a previous condition or new logic will not work

src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -31,23 +31,44 @@ public void SaveState(byte[] state)
...
+                    var data = LoadState();
+                    SaveState(data.Concat(receipt.Hash.ToBytes()).ToArray());
                 }
             }
-            else
+            catch (Exception e)
             {
-                var data = LoadState();
-                SaveState(data.Concat(receipt.Hash.ToBytes()).ToArray());
+                Console.WriteLine($"exception occured while decoding: {e}");
```
sgladkov, line 71: Don't use write to console in Core code,  we have sofisticated log processing functionality,  this output can create an issue. We don't need special output for this case really,  it is normal to has exception here,  no need in warnings.
Also we need to check To and from fields here.

src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -31,23 +31,44 @@ public void SaveState(byte[] state)
...
-                if (res!.Equals(receipt.Transaction.To))
+                object[] decodedRes = {};
+                
+                foreach (var signature in signatures)
+                {
+                    decodedRes = decoder.Decode(signature);
+                    break;
+                }
+                
+                if (decodedRes.Length >= 5)
```
sgladkov, line 52: Why we check 4-th place here only? As far as I can see from our signatures it should be decodedRes[0] or decodedRes[1] (we can check every member really  if it has address type)

sgladkov (changes requested):

sgladkov:
Add some test for this functionality, please. For example test local repository for several sample transactions of different types

src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
dushyantb: 4th place because it is having the address

sgladkov: this array hasn't 4 members for all of our signatures,  it has 2 members 3 of them and 3 members for  1 signature,  this code doesn't work as far as I can see. Search other usage of the Decode method in the code.

src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -49,13 +49,10 @@ public void TryAddTransaction(TransactionReceipt receipt)
                     break;
                 }
                 
-                if (decodedRes.Length >= 5)
+                if (decodedRes[0] is UInt160 decodedAddr && decodedAddr.Equals(receipt.Transaction.To))
```
sgladkov, line 52: Now you are testing only first argument. We have signatures where we should check not only first argument (transferFrom(address,address,uint256)),  this case will be missed. Just check all arguments please,  we can add some signatures in the future .

sgladkov (changes requested):

src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -33,32 +33,47 @@ public void TryAddTransaction(TransactionReceipt receipt)
...
-                
+
                 var decoder = new ContractDecoderLtr(receipt.Transaction.Invocation.ToArray());
-                object[] decodedRes = {};
-                
+                object[] decodedRes = { };
+
                 foreach (var signature in signatures)
                 {
                     decodedRes = decoder.Decode(signature);
```
sgladkov, line 49: You are checking only first signature really,  then break is called

src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -33,32 +33,47 @@ public void TryAddTransaction(TransactionReceipt receipt)
...
                     break;
                 }
-                
-                if (decodedRes[0] is UInt160 decodedAddr && decodedAddr.Equals(receipt.Transaction.To))
+
+                if (decodedRes.Length == 3)
                 {
-                    var temp = LoadState();
-                    SaveState(temp.Concat(receipt.Hash.ToBytes()).ToArray());     
+                    if (decodedRes[0] is UInt160 fromAddr && fromAddr.Equals(receipt.Transaction.From) &&
```
sgladkov, line 55: 1)this checks should be inside the cycle with signatures
2)why do you compare receipt.Transaction params with decoded script call arguments? Your goal is to fiind local address among this arguments,  transaction params are not related to this search

sgladkov (changes requested):
Please,  remove the commits that not related to this pull request from the branch,  it is very difficult to review and to merge this pull request. And add the tests to be sure the code is working

src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -31,57 +31,55 @@ public void SaveState(byte[] state)
...
                 }
-                else
+                catch (Exception e)
                 {
-                    if (_watchAddresses.Count(addr =>
-                        addr.Equals(receipt.Transaction.To) || addr.Equals(receipt.Transaction.From)) <= 0) return;
-
-                    var data = LoadState();
-                    SaveState(data.Concat(receipt.Hash.ToBytes()).ToArray());
+                    Console.WriteLine($"exception occured while decoding: {e}");
```
sgladkov, line 80: Don't use Console.WriteLine please in Core code,  we don't need any special notification here,  every transaction will throw at least 3 exception here,  too many useless console output or log records

sgladkov (changes requested):

PR #119, opened 2021-11-18
added hardfork params, cycleDuration and validatorCount in config file

HMAshiqulIslam:


src/Lachain.Console/TrustedKeygen.cs
```diff
@@ -38,22 +42,24 @@ internal class Config
             [JsonProperty("vault")] public VaultConfig Vault { get; set; }
             [JsonProperty("storage")] public StorageConfig Storage { get; set; }
             [JsonProperty("blockchain")] public BlockchainConfig Blockchain { get; set; }
+            [JsonProperty("hardfork")] public HardforkConfig Hardfork { get; set; }
+            [JsonProperty("version")] public VersionConfig ConfigVersion { get; set; }
```
HMAshiqulIslam, line 46: do you think it's better to name "ConfigVersion" to "Version" keeping it consistent with the rest ?

sgladkov: It is conflicting with other vars in this case

HMAshiqulIslam: which one ? I can't see any other vars with the same name.

PR #120, opened 2021-11-19
feature larc_history_api

dushyantb:


src/Lachain.Storage/Repositories/LocalTransactionRepository.cs
```diff
@@ -31,11 +35,55 @@ public void SaveState(byte[] state)
...
+                    {
+                        if (_watchAddresses.Count(addr =>
+                            decodedRes[0] is UInt160 fromAddr && fromAddr.Equals(addr)) <= 0) continue;
+
+                        var temp = LoadState();
+                        SaveState(temp.Concat(receipt.Hash.ToBytes()).ToArray());
+                    }
+                    else
+                    {
+                        if (_watchAddresses.Count(addr =>
```
sgladkov, line 75: This code will be executed really rarely. For the most part of transactions (for example,  transactions without contract calls) we get exception during decoder.Decode(signature) call and goes to the next signature without checking this condition. This code should be outside the try-catch block (and after the cycle really).

sgladkov (changes requested):

sgladkov (approved):

PR #122, opened 2021-11-22
Fix SyncBlock request processing

sgladkov:


src/Lachain.Core/Network/MessageHandler.cs
```diff
@@ -104,25 +104,36 @@ private void OnPingReply(object sender, (PingReply reply, ECDSAPublicKey publicK
             using var timer = IncomingMessageHandlingTime.WithLabels("SyncBlocksRequest").NewTimer();
             Logger.LogTrace("Start processing SyncBlocksRequest");
             var (request, callback) = @event;
-            var reply = new SyncBlocksReply
+            if (request.ToHeight > request.FromHeight)
```
HMAshiqulIslam, line 107: Should the comparator be >= ? 
I think the range is inclusive in both end.

sgladkov: Good catch.)

PR #125, opened 2021-11-24
Fixed a consensus unit test of binary broadcast

HMAshiqulIslam:
(1) The unit test assumed that, **All the process will return the same bin_values**. But this is only true after the binary broadcast finishes its execution. In an asynchronous setting, it's never possible to know if the execution has finished its execution.
Ref: [https://hal.inria.fr/hal-00944019v2/document](url) : Page 5 - Remark

(2) Our binary broadcast executes till **step 05** shown in **figure 2** of [https://hal.inria.fr/hal-00944019v2/document](url) 
It's not guaranteed that, if at least (f+1) process gives the same binary input, it will return that binary value (this is also because incompleteness of **step 03**). 

But the reverse is true: 
**If a correct process returns a binary value, that was definitely proposed by at least (f+1) correct processes.**

### DIFF
--- a/test/Lachain.ConsensusTest/BinaryBroadcastTest.cs
+++ b/test/Lachain.ConsensusTest/BinaryBroadcastTest.cs
@@ -129,7 +129,7 @@ namespace Lachain.ConsensusTest
         }
 
         [Test]
-        [Repeat(10)]
+        [Repeat(1000)]
         public void TestRandomValues()
         {
             var n = _rnd.Next(1, 10);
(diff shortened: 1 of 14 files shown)


test/Lachain.ConsensusTest/BinaryBroadcastTest.cs
```diff
@@ -129,7 +129,7 @@ public void TestRandomFailures_7_2()
         }
 
         [Test]
-        [Repeat(10)]
+        [Repeat(1000)]
```
sgladkov, line 132: decrease repeat count please (return it to old value),  1000 is too long for unit tests. We will add this trest to the external integratioin tests in future

sgladkov (changes requested):

test/Lachain.ConsensusTest/BinaryBroadcastTest.cs
sgladkov:
```suggestion
        [Repeat(10)]
```

PR #126, opened 2021-11-24
Catch possible exception in eth_getBlockByNumber

sgladkov:


src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
```diff
@@ -69,15 +69,26 @@ public class BlockchainServiceWeb3 : JsonRpcService
 
             ulong gasUsed = 0;
             var txArray = new JArray();
-            if(block.TransactionHashes.Count>0)
-            {
-                var txs = block!.TransactionHashes
+            if (block.TransactionHashes.Count <= 0) 
+                return Web3DataFormatUtils.Web3Block(block!, gasUsed, txArray);
+            var txs = block!.TransactionHashes
```
HMAshiqulIslam, line 74: Should we have a try catch block for this statement as well ?

Arghya115, line 74: Can we also put this line in try-catch block? I've a suspicion, this line might also be problematic.

HMAshiqulIslam (approved):

PR #133, opened 2021-11-26
used Web3Number instead of Web3Data to return tx details

dushyantb:


src/Lachain.Core/RPC/HTTP/Web3/Web3DataFormatUtils.cs
```diff
@@ -202,13 +207,13 @@ public static IHashTrieNode NodeFromJson(JObject jsonNode)
                 ["gas"] = Web3Number(receipt.Transaction.GasLimit),
                 ["gasPrice"] = Web3Number(receipt.Transaction.GasPrice),
                 ["hash"] = Web3Data(receipt.Hash),
-                ["input"] = Web3Data(receipt.Transaction.Invocation),
+                ["input"] = Web3Number(receipt.Transaction.Invocation),
```
sgladkov, line 210: Are you sure that this field is a number?  Invocation is a piece  of the bytecode really,  it can't be used as a number

src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
```diff
@@ -69,16 +69,29 @@ public class BlockchainServiceWeb3 : JsonRpcService
 
             ulong gasUsed = 0;
             var txArray = new JArray();
-            if(block.TransactionHashes.Count>0)
+
```
sgladkov, line 72: Remove changes in this file,  it overwrites the last changes

src/Lachain.Core/RPC/HTTP/Web3/Web3DataFormatUtils.cs
```diff
@@ -29,6 +29,11 @@ public static string Web3Number(ulong x)
         {
             return x.ToHex(false);
         }
+        
+        public static string Web3Number(IEnumerable<byte> x)
+        {
+            return x.ToHex();
```
sgladkov, line 35: Are you sure it is a correct convertation? It should be similar to the function for UInt256 type, as far as I remember

sgladkov (changes requested):

PR #136, opened 2021-11-29
fixed invalid tx issue from blocks

HMAshiqulIslam:


src/Lachain.Core/Consensus/BlockProducer.cs
```diff
@@ -66,7 +66,7 @@ public IEnumerable<TransactionReceipt> GetTransactionsToPropose(long era)
                 hashesTaken = new UInt256[]{};
                 return null;
             }
-            var txsGot = _blockSynchronizer.WaitForTransactions(txHashes, TimeSpan.FromDays(1)); // TODO: timeout?
+            var txsGot = _blockSynchronizer.WaitForTransactions(txHashes, TimeSpan.FromDays(1), out List<TransactionReceipt> receipts); // TODO: timeout?
```
sgladkov, line 69: I think we should decrease this timeout now.  I think 1 hour should be enough.
```suggestion
            var txsGot = _blockSynchronizer.WaitForTransactions(txHashes, TimeSpan.FromHours(1), out List<TransactionReceipt> receipts); // TODO: timeout?
```

sgladkov (approved):

PR #137, opened 2021-12-01
Testnet api issue

Arghya115:
Added some logs to track problematic transaction. We can remove them once the problem is solved.

### DIFF
--- a/src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
+++ b/src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
@@ -103,6 +103,8 @@ namespace Lachain.Core.RPC.HTTP.Web3
                 Logger.LogWarning($"txs {txs}");
                 foreach (var tx in txs)
                 {
+                    if (tx is null) 
+                         continue;
                     Logger.LogWarning($"tx {tx.Hash.ToHex()} {tx.GasUsed} {tx.Status} {tx.IndexInBlock}");
                 }
             }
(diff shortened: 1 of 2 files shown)


HMAshiqulIslam (approved):

src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
```diff
@@ -96,10 +103,9 @@ public class BlockchainServiceWeb3 : JsonRpcService
                 Logger.LogWarning($"txs {txs}");
                 foreach (var tx in txs)
                 {
-                    Logger.LogWarning($"tx {tx.Hash} {tx.GasUsed} {tx.Status} {tx.IndexInBlock}");
+                    Logger.LogWarning($"tx {tx.Hash.ToHex()} {tx.GasUsed} {tx.Status} {tx.IndexInBlock}");
```
sgladkov, line 106: We need to check if tx is not empty here
```suggestion
                    if (tx is null) 
                         continue;
                    Logger.LogWarning($"tx {tx.Hash.ToHex()} {tx.GasUsed} {tx.Status} {tx.IndexInBlock}");
```

sgladkov (approved):

sgladkov:
> ```
>                 if(tx is null)
>                 {
>                     Logger.LogWarning($"Transaction not found in DB {txHash.ToHex()}");
>                 }
>                 else txs.Add(tx);
> ```
> 
> So null tx is not being added to the txs.

Yes,  you are right,  my fail.) But my addition doesn'tr break anything,  so we can leave it for simplicity.)

PR #142, opened 2021-12-03
Process uncatched exceptions during the tx execution

sgladkov:


Arghya115 (approved):

src/Lachain.Core/Blockchain/Operations/BlockManager.cs
```diff
@@ -350,7 +350,16 @@ private void BlockPersisted(Block block)
...
-                    var result = _transactionManager.Execute(block, receipt, snapshot);
+                    OperatingError result = OperatingError.Ok;
+                    try
+                    {
+                        result = _transactionManager.Execute(block, receipt, snapshot);
+                    }
+                    catch (Exception e)
+                    {
+                        Logger.LogWarning($"Exception during tx execution: {e}");
+                        result = OperatingError.InvalidContract;
```
HMAshiqulIslam, line 361: Do you think this invalidContract could be misleading ? We know nothing about the issue here..
We can instead say unknownError. What do you think ?

HMAshiqulIslam (approved):

src/Lachain.Core/Blockchain/Operations/BlockManager.cs
sgladkov: it is something wrong with contract if we have an exception here. And we haven't unknown error for this type,  I think it is a good practice

PR #150, opened 2021-12-15
added pending state getbalance

kunalgoyal9:


src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -39,8 +39,42 @@ public class AccountServiceWeb3 : JsonRpcService
...
-            var availableBalance = GetSnapshotByTag(tag)!.Balances.GetBalance(addressUint160);
-            return Web3DataFormatUtils.Web3Number(availableBalance.ToWei().ToUInt256());
+
+            if (tag == "pending")
+            {
+                //Extract all the transactions from the pool
+                //Extract the transactions where 'to' is this address(let's say type-1)
+                //Extract the transactions where 'from' is this address(let's say type-2)
+                //Then you have to simulate execution of all these transactions in the(ReceiptCompare Order).
+                //Then you can find the pending balance of the address.
```
HMAshiqulIslam, line 49: The comments are a bit misleading. We should comment the version we are considering.

HMAshiqulIslam (changes requested):
Let's keep a check for the nonce. 
(1) First get the transaction count from the latest snapshot. 
(2) Then sort all the transactions from "address" in nonce order
(3) Then try to do calculation and before calculation, check if the nonce is correct, if it's correct, add 1 to the nonce, otherwise skip that transaction.

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -39,8 +40,60 @@ public class AccountServiceWeb3 : JsonRpcService
...
+            {
+                // Get all transaction from pool
+                var txpool = this._transactionPool;
+                var transactions = txpool.Transactions;
+
+                List<TransactionReceipt> tx_Receipts = new List<TransactionReceipt>();
+                
+                foreach(var tx in transactions)
+                {
+                    tx_Receipts.Add(tx.Value);
```
HMAshiqulIslam, line 54: We use camelCase for local variables. Let's stay consistent.

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -39,8 +40,60 @@ public class AccountServiceWeb3 : JsonRpcService
...
         {
             var addressUint160 = address.HexToUInt160();
-            var availableBalance = GetSnapshotByTag(tag)!.Balances.GetBalance(addressUint160);
-            return Web3DataFormatUtils.Web3Number(availableBalance.ToWei().ToUInt256());
+
+            if (tag == "pending")
+            {
+                // Get all transaction from pool
+                var txpool = this._transactionPool;
+                var transactions = txpool.Transactions;
```
HMAshiqulIslam, line 48: Should we change this to just "_transactionPool.Transactions"

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -39,8 +40,60 @@ public class AccountServiceWeb3 : JsonRpcService
...
+                {
+                    tx_Receipts.Add(tx.Value);
+                }
+
+                // Sort on the basis of nonce
+                tx_Receipts = tx_Receipts.OrderBy(receipt => receipt, new ReceiptComparer()).ToList();
+
+                // Get current address nonce
+                var transactionRepository = _stateManager.CurrentSnapshot.Transactions;
+                var curr_nonce = transactionRepository.GetTotalTransactionCount(address.HexToBytes().ToUInt160());
```
HMAshiqulIslam, line 62: We can use addressUint160 instead of reformatting again.

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -39,8 +40,60 @@ public class AccountServiceWeb3 : JsonRpcService
...
+
+                List<TransactionReceipt> tx_Receipts = new List<TransactionReceipt>();
+                
+                foreach(var tx in transactions)
+                {
+                    tx_Receipts.Add(tx.Value);
+                }
+
+                // Sort on the basis of nonce
+                tx_Receipts = tx_Receipts.OrderBy(receipt => receipt, new ReceiptComparer()).ToList();
```
HMAshiqulIslam, line 58: Should we consider sorting after we have filterred the necessary transactions ? 
Pool may be very large, it's not a good idea to sort them in each api call.
On the other hand, pending transactions from a particular address will be very small.

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -39,8 +40,60 @@ public class AccountServiceWeb3 : JsonRpcService
...
+                    {
+                        // Executing the transaction
+                        var gasp = new Money(tx.Transaction.GasPrice);
+                        var gasl = new Money(tx.Transaction.GasLimit);
+                        var txamnt = new Money(tx.Transaction.Value);
+
+                        availableBalance = availableBalance - txamnt  - (gasl * gasp);
+
+                        // Check if balance is less than 0
+                        if(availableBalance < Money.Parse("0"))
```
HMAshiqulIslam, line 81: this is not true. 
Before substracing from the balance, we need to check if this is okay to execute. If yes, then we substract, otherwise we just skip it. 

for example, let's say address A has 100 LA. 
If it sends 3 tx sending 200 LA, 20 LA, 20 LA. 
then, the first will be a failed transaction (we still have to increase the nonce and substract the gas). 
then it will execute the second and third tx safely. 

Address's pending balance should be = 100 - 20 - 20 - gas1 - gas2 - gas3

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -39,8 +40,60 @@ public class AccountServiceWeb3 : JsonRpcService
...
+                        {
+                            return Web3DataFormatUtils.Web3Number(Money.Parse("0").ToWei().ToUInt256());
+                        }
+                    }
+
+                    curr_nonce += 1;
```
HMAshiqulIslam, line 87: This nonce update should be inside the If statement. 
We should only increase the nonce if the transaction's from == address

HMAshiqulIslam (changes requested):
Please take a look at the comments.

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
kunalgoyal9: done

test/Lachain.CoreTest/RPC/HTTP/Web3/AccountServiceWeb3Test.cs
```diff
@@ -115,6 +117,19 @@ public void Test_GetBalance()
             Assert.AreEqual(balNew, "0x115557b419c5c1f3fa018400000000");
         }
 
+        [Test]
+        // Changed GetBalance to public
+        public void Test_GetBalance_pending()
```
HMAshiqulIslam, line 122: Let's also keep the function name consistent with the rest. 
"Test_GetBalancePending"

test/Lachain.CoreTest/RPC/HTTP/Web3/AccountServiceWeb3Test.cs
```diff
@@ -115,6 +117,19 @@ public void Test_GetBalance()
...
+        {
+            var address = "0x2605c1ad496f428ab2b700edd257f0a378f83750";
+
+            Execute_dummy_transaction(false);
```
HMAshiqulIslam, line 126: These types of function names are wrriten in BlockCase. 
Let's change it to 
"ExecuteDummyTransaction"

HMAshiqulIslam (approved):

PR #164, opened 2021-12-29
Commit made atomic

Arghya115:


src/Lachain.Storage/State/StateManager.cs
```diff
@@ -21,8 +20,9 @@ public class StateManager : IStateManager
         private readonly ISnapshotManager<IValidatorSnapshot> _validatorManager;
 
         private readonly Mutex _globalMutex = new Mutex(false);
+        public IRocksDbContext _dbContext;
```
HMAshiqulIslam, line 23: should it be public ?

src/Lachain.Storage/StorageState.cs
```diff
@@ -104,12 +104,12 @@ public ulong TryDelete(byte[] key, out byte[]? value)
...
 
         [MethodImpl(MethodImplOptions.Synchronized)]
-        public ulong Commit()
+        public ulong Commit(RocksDbAtomicWrite batch)
         {
-            using var tx = _repositoryManager.CreateTransaction();
-            _trieMap.Checkpoint(CurrentVersion, tx);
-            _repositoryManager.SetState(CurrentVersion, tx);
-            tx.Commit();
+        //    using var tx = _repositoryManager.CreateTransaction();
```
HMAshiqulIslam, line 109: what does this comments mean ?

src/Lachain.Storage/State/StateManager.cs
```diff
@@ -143,13 +145,20 @@ public void Rollback()
...
+            _contractManager.Commit(batch);
+            _storageManager.Commit(batch);
+            _transactionManager.Commit(batch);
+            _blockManager.Commit(batch);
+            _eventManager.Commit(batch);
+            _validatorManager.Commit(batch);
+            batch.Commit();
+        }
+
+        public void Commit(RocksDbAtomicWrite batch)
```
HMAshiqulIslam, line 159: why do we need this ?

HMAshiqulIslam (changes requested):

src/Lachain.Storage/State/StateManager.cs
Arghya115: StateManager inherits from IStateManager and IStateManager inherits from ISnapshotManager<IBlockchainSnapshot>, now SnapshotManager also inherits from ISnapshotManager. I needed to put Commit(batch) in the ISnapshotManager and as a result I also had to put it in StateManager. This looks really stupid but I can't think of any other way without massive change in the architecture.

src/Lachain.Storage/StorageState.cs
Arghya115: Removing this now.

src/Lachain.Storage/State/StateManager.cs
Arghya115: Not sure if there is any problem with public, but keeping it private is better idea. Changing it now.

src/Lachain.Storage/State/StateManager.cs
HMAshiqulIslam: maybe we can add some comments or throw exception inside this method to make sure none uses it accidentally.

sgladkov (approved):

PR #167, opened 2022-01-05
Added topics for external contract

tbssajal:


sgladkov (commented):
Have you tested if hardfork required here?

PR #169, opened 2022-01-10
Web3 api

tbssajal:


src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
```diff
@@ -568,9 +523,81 @@ public string GetNetworkGasPrice()
...
+            ulong? gasPriceToUse = null;
+            if(!(gasPrice is null)) gasPriceToUse  = ((string)gasPrice!).HexToUlong();
+            
+            byte[]? byteCode = null;
+            if(!(data is null)) byteCode = ((string) data!).HexToBytes();
+            
+            
+            var config = _configManager.GetConfig<VaultConfig>("vault") ??
+                         throw new Exception("No 'vault' section in config file");
+            string password = config.Password!;
```
sgladkov, line 564: It is a wrong way. You can't access a config password here,  in this case anyone can send a transaction signed with this node and get access to any node money this way. Instead you should check if _privateWallet is unlocked and sign tx in this case only (throw an exception in other case). Also add this method to the private methods list,

sgladkov (changes requested):

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
tbssajal: I added "eth_sendTransaction" and "eth_signTransaction" in private methods list.

sgladkov: You still can't access config password in this case. Check _privateWallet is unlocked already during the call,  don't unlock it yourself

tbssajal: done

sgladkov (approved):

PR #170, opened 2022-01-13
Pull Request from branch [feature/ecdsa-auth]

kazunetakeda25:


src/Lachain.Console/TrustedKeygen.cs
```diff
@@ -135,7 +135,7 @@ internal class Config
                     Hosts = new[] {"+"},
                     Port = basePort,
                     MetricsPort = (ushort) (basePort + 1),
-                    ApiKey = "asdasdasd",
+                    ApiKey = "0x2e917846fe7487a4ea3a765473a3fc9b2d9227a4d312bc77fb9de357cf73d7e52b771d537394336e9eb2cb4838138f668f4bd7d8cf7e04d9242a42c71b99f166",
```
sgladkov, line 138: We meed to generate random key here

src/Lachain.Console/TrustedKeygen.cs
```diff
@@ -275,7 +275,7 @@ internal class Config
                     Hosts = new[] {"+"},
                     Port = (ushort) (basePort + i),
                     MetricsPort = (ushort) (basePort + n + i),
-                    ApiKey = "asdasdasd",
+                    ApiKey = "0x2e917846fe7487a4ea3a765473a3fc9b2d9227a4d312bc77fb9de357cf73d7e52b771d537394336e9eb2cb4838138f668f4bd7d8cf7e04d9242a42c71b99f166",
```
sgladkov, line 278: We need to generate random key here

src/Lachain.Core/RPC/HTTP/HttpService.cs
```diff
@@ -147,19 +159,36 @@ private bool _Handle(HttpListenerContext context)
...
+            }            
 
             JsonRpcProcessor.Process(async);
             return true;
         }
 
-        private bool _CheckAuth(JObject body, HttpListenerContext context)
+        private bool _CheckAuth(JObject body, HttpListenerContext context, string sig, string bodyPlainText)
         {
+			if (string.IsNullOrEmpty(_apiKey))
```
sgladkov, line 170: We don't need to check it for every request,  it is better to check it only once during HttpService start

src/Lachain.Console/TrustedKeygen.cs
sgladkov: And print associated private key to logs to have ability to save it for future use in client

sgladkov (changes requested):

src/Lachain.Core/RPC/HTTP/HttpService.cs
kazunetakeda25: Removed

src/Lachain.Console/TrustedKeygen.cs
kazunetakeda25: Fixed in commit #170 8421408a

src/Lachain.Console/TrustedKeygen.cs
kazunetakeda25: Fixed in commit #170 8421408a

PR #175, opened 2022-01-14
Feature/sanitize low gas transactions

HMAshiqulIslam:


src/Lachain.Core/Blockchain/Pool/TransactionPool.cs
```diff
@@ -122,9 +148,9 @@ public OperatingError Add(TransactionReceipt receipt, bool notify = true)
             if (_transactions.ContainsKey(receipt.Hash))
                 return OperatingError.AlreadyExists;
             /* verify transaction before adding */
-            if (!(GetNextNonceForAddress(receipt.Transaction.From) >= receipt.Transaction.Nonce && TxNonceValid(receipt)))
+            if (!(GetNextNonceForAddress(receipt.Transaction.From) == receipt.Transaction.Nonce))
```
sgladkov, line 151: Does it mean that we can't add tx to pool if its nonce is bigger then next nonce value?

src/Lachain.Core/Blockchain/Pool/TransactionPool.cs
```diff
@@ -350,17 +347,27 @@ public IReadOnlyCollection<TransactionReceipt> Peek(int txsToLook, int txsToTake
...
-                    continue;
-                _relayQueue.Add(receipt);
+                Logger.LogError($"_transactions count: {_transactions.Count()} != _transactionsQueue: {_transactionsQueue.Count()}");
+                throw new Exception("transactions and transactionQueue should be of same size");
+            }
+
+            // at this point nonce-calculator and transaction should have the 
+            // same set of transactions
+
+            if(_transactions.Count() != _nonceCalculator.Count())
```
sgladkov, line 367: Should we compare tx hashes too if they should contain the same set?

PR #178, opened 2022-01-28
in-memory cache implementation for BlockManager.GetByHeight

kazunetakeda25:
Files changed:
- TrustedKeygen.cs (Modified)
- BlockManager.cs (Modified)
- CacheConfig.cs (Added)
- CaheOptions.cs (Added)

### DIFF
--- a/src/Lachain.Core/Blockchain/Operations/BlockManager.cs
+++ b/src/Lachain.Core/Blockchain/Operations/BlockManager.cs
@@ -166,14 +166,14 @@ namespace Lachain.Core.Blockchain.Operations
         {
             try
             {
-                /* try to block from cache */
+                /* try to get block from cache */
                 if (_heightCache.TryGetValue(blockHeight, out var block))
                     return block;
 
                 var newBlock = _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);
 
                 /* if it reach cache size limit, remove the oldest one */
-                if (_heightCacheQueue.Count == _blockSizeLimit)
+                if (_heightCacheQueue.Count == _blockSizeLimit && _heightCacheQueue.Count > 0)
                 {
                     /* remove from queue */
                     var oldestKey = _heightCacheQueue.Dequeue();
--- a/src/Lachain.Core/Blockchain/Operations/BlockManager.cs
+++ b/src/Lachain.Core/Blockchain/Operations/BlockManager.cs
@@ -31,6 +31,8 @@ namespace Lachain.Core.Blockchain.Operations
     {
         private static readonly ILogger<BlockManager> Logger = LoggerFactory.GetLoggerForClass<BlockManager>();
         private static readonly ICrypto Crypto = CryptoProvider.GetCrypto();
+        private IDictionary<ulong, Block?> _heightCache
+            = new Dictionary<ulong, Block?>();
 
         private static readonly Counter BlockExecCounter = Metrics.CreateCounter(
             "lachain_block_exec_count",
@@ -150,7 +152,36 @@ namespace Lachain.Core.Blockchain.Operations
         [MethodImpl(MethodImplOptions.Synchronized)]
         public Block? GetByHeight(ulong blockHeight)
         {
-            return _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);
+            try
+            {
+                /* default cache size limit is 100 items */
+                var sizeLimit = 100;
+                var cacheConfig = _configManager.GetConfig<CacheConfig>("cache");                
+                if (cacheConfig != null && cacheConfig.BlockHeight != null && cacheConfig.BlockHeight.SizeLimit != null)
+                {
+                    sizeLimit = cacheConfig.BlockHeight.SizeLimit.Value; 
+                }
+
+                /* try to block from cache */
+                if (_heightCache.TryGetValue(blockHeight, out var block))
+                    return block;
+
+                var newBlock = _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);
+
+                /* if it reach cache size limit, remove the oldest one */
+                if (_heightCache.Count == sizeLimit)
+                    _heightCache.Remove(_heightCache.First());
+
+                /* then add new key to cache */
+                _heightCache.Add(blockHeight, newBlock);
+
+                return newBlock;
+            }
+            catch (Exception)
+            {
+                return null;
+            }
+
         }
 
         [MethodImpl(MethodImplOptions.Synchronized)]
--- a/src/Lachain.Core/Blockchain/Operations/BlockManager.cs
+++ b/src/Lachain.Core/Blockchain/Operations/BlockManager.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
@@ -31,8 +31,8 @@ namespace Lachain.Core.Blockchain.Operations
     {
         private static readonly ILogger<BlockManager> Logger = LoggerFactory.GetLoggerForClass<BlockManager>();
         private static readonly ICrypto Crypto = CryptoProvider.GetCrypto();
-        private SortedDictionary<ulong, Block?> _heightCache
-            = new SortedDictionary<ulong, Block?>();
+        private IDictionary<ulong, Block?> _heightCache;
+        private Queue<ulong> _heightCacheQueue;
 
         private static readonly Counter BlockExecCounter = Metrics.CreateCounter(
             "lachain_block_exec_count",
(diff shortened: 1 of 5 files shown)


src/Lachain.Core/Blockchain/Operations/BlockManager.cs
```diff
@@ -150,7 +152,36 @@ public Block LatestBlock()
         [MethodImpl(MethodImplOptions.Synchronized)]
         public Block? GetByHeight(ulong blockHeight)
         {
-            return _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);
+            try
+            {
+                /* default cache size limit is 100 items */
+                var sizeLimit = 100;
```
sgladkov, line 158: It is better to read cache limit only once in constructor instead of reading it every time block is requested

sgladkov (changes requested):
If you want to add some new value to config add the corresponding automatic update  with default value for it in ConfigManager.cs

src/Lachain.Core/Blockchain/Operations/BlockManager.cs
kazunetakeda25: Okay

src/Lachain.Core/Blockchain/Operations/BlockManager.cs
```diff
@@ -150,7 +161,28 @@ public Block LatestBlock()
...
+            {
+                /* try to block from cache */
+                if (_heightCache.TryGetValue(blockHeight, out var block))
+                    return block;
+
+                var newBlock = _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);
+
+                /* if it reach cache size limit, remove the oldest one */
+                if (_heightCache.Count == _blockSizeLimit)
+                    _heightCache.Remove(_heightCache.First());
```
HMAshiqulIslam, line 174: This is actually O(# of elements). So, it would be very slow and not scalable if we want to store let's say 1 million items in cache. What do you think about it ?

kazunetakeda25: so block size limit is set as 100 by default.

HMAshiqulIslam: I mean - we should make the erase faster than O(# of elements). Can we use a sortedSet (https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.sortedset-1?view=net-6.0) instead of a dictionary ?

sgladkov: Looks like Dictionary.Remove has O(1) efficiency,  https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.remove?view=net-5.0 : This method approaches an O(1) operation.

HMAshiqulIslam: Actually the culprint the first(). It's a linq expression which is O(n). Dictionary itself has no first() method.

kazunetakeda25: Hi @HMAshiqulIslam how about using ElementAt(0)?

HMAshiqulIslam: A dictionary is unordered. I don't think there is any way to find the minimum key without iterating over all the keys. 
So elementAt(0) should also work in O(n).

kazunetakeda25: @HMAshiqulIslam, @sgladkov, Dictionary in principle has a faster lookup with O(1) while the lookup performance of a List is an O(n) operation.
if we have to use SortedSet, we have to use ValueKeyPair, so for our case, SortedDictonary is better. but there is no way to get the element at 0, I will go with elementAt(0) or First().. looks same for me.
Please review my recent commit.

HMAshiqulIslam: @kazunetakeda25 
I think we still have the same issue. _heightCache.First().Key
What does First() really return ? How fast is this operation ?

kazunetakeda25: @HMAshiqulIslam Please share me a better option if you have any... I have no idea...

kazunetakeda25: I think the removal of the first element is always O(n).
Maybe we should think different approach... Please let me know @sgladkov @HMAshiqulIslam

kazunetakeda25: Are queue kinds helpful then?

HMAshiqulIslam: I am bit confused if for a sortedDictionary, .First() is O(n) or O(logN). 
One other approach that we can take is as follows: 
we keep the keys in a sortedSet and (key, value) to another dictionary. 
Then, we can do sortedSet.Min to find the minimum key and erase them from both the dictionary and the sortedSet.

HMAshiqulIslam: Also, for this scenerio, queues should also work. We can store the keys in a queue and a separate dictionary for (key, value) pairs and erase the key from the front of the queue. 
What do you think ? @kazunetakeda25

kazunetakeda25: Looks good to me @HMAshiqulIslam

kazunetakeda25: Hi @sgladkov @HMAshiqulIslam Please review my recent new commit.
I used Queue and all methods i used are O(1)

HMAshiqulIslam: This looks good to me. One thing to note : do you think we should implement an isolated generic FIFO cache and make use of this in the blockManager ? We will be using this cache a lot later in different methods.

sgladkov: We can move buffer to separate abstraction in the next PR where we will use it for other method. Let's implement it in several iterations.

sgladkov (dismissed):
LGTM

src/Lachain.Core/Blockchain/Operations/BlockManager.cs
kazunetakeda25: I am currently working on it already. @sgladkov @HMAshiqulIslam

src/Lachain.Core/Blockchain/Operations/BlockManager.cs
```diff
@@ -150,7 +164,35 @@ public Block LatestBlock()
...
+                if (_heightCache.TryGetValue(blockHeight, out var block))
+                    return block;
+
+                var newBlock = _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);
+
+                /* if it reach cache size limit, remove the oldest one */
+                if (_heightCacheQueue.Count == _blockSizeLimit)
+                {
+                    /* remove from queue */
+                    var oldestKey = _heightCacheQueue.Dequeue();
```
HMAshiqulIslam, line 179: Do you think this will fail for : 
_blockSizeLimit = 0 

Should we allow _blockSizeLimit = 0 ? I think - yes - as this would mean that no caching is being done.

kazunetakeda25: Hi @HMAshiqulIslam 
_blockSizeLimit = 0 means no caching enabled.
But _blockSizeLimit is being set in config.json. And the default value is 100. So caching is enabled by default.

HMAshiqulIslam: I meant : The function will probably not work when _blockLimitSize = 0. It will always return null. Can you please test it ?

kazunetakeda25: Hi @HMAshiqulIslam I will check it

kazunetakeda25: Please check my latest commit @sgladkov @HMAshiqulIslam

HMAshiqulIslam: I think - when we set _blockSizeLimit = 0, the items in the cache will never be removed due to this equality 
check: **_heightCacheQueue.Count == _blockSizeLimit**
You can test it locally. I think - every blocks will be stored in the cache which will increase the memory usage gradually.

kazunetakeda25:
```
if (_blockSizeLimit == 0)
{
    return _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);
}

try
{
    /* try to get block from cache */
    if (_heightCache.TryGetValue(blockHeight, out var block))
        return block;

    var newBlock = _stateManager.LastApprovedSnapshot.Blocks.GetBlockByHeight(blockHeight);

    /* if it reach cache size limit, remove the oldest one */
    if (_heightCacheQueue.Count == _blockSizeLimit)
    ...
```
Looks good now?

HMAshiqulIslam: Looks perfect now. Thanks.

HMAshiqulIslam (approved):

sgladkov (approved):

PR #180, opened 2022-02-03
Fix attendance after the break in validator status

sgladkov:
The issue arises after the node was validator for some time and stored Attendance info in database,  than it stopped to be validator for some time,  and when it becomes validator again it can't load correct Attendance object,  the object it gets can't be changed with the data from the current cycle


HMAshiqulIslam (approved):
LGTM

PR #182, opened 2022-02-09
Fixed in-memory cache for blocks

HMAshiqulIslam:
(1) We add to the cache only after a new block is persisted. It's important that we do not store any block during its emulation phase as it's not final yet and is subject to change. 
(2) It's also important that null blocks are not stored in the cache.


sgladkov (approved):

kazunetakeda25 (approved):
Looks good to me.

PR #183, opened 2022-02-10
Allow validator to return the stake

sgladkov:
Validator itself could have opportunity to stop the work and return the stake, that's why it should have permission to request stake withdraw


HMAshiqulIslam (approved):
LGTM

PR #189, opened 2022-02-11
Feature/eth sign

kazunetakeda25:
Added `eth_sign` method

### DIFF
--- a/src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
+++ b/src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
@@ -192,7 +192,7 @@ namespace Lachain.Core.RPC.HTTP.Web3
             }
 
             var addressUInt160 = (address!).HexToUInt160();
-            if (addressUInt160.Equals(_privateWallet.EcdsaKeyPair.PublicKey.GetAddress()))
+            if (!addressUInt160.Equals(_privateWallet.EcdsaKeyPair.PublicKey.GetAddress()))
             {
                 throw new Exception("address is invalid");
             }
--- a/src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
+++ b/src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
@@ -30,6 +30,8 @@ namespace Lachain.Core.RPC.HTTP.Web3
         private readonly ITransactionPool _transactionPool;
         private readonly IPrivateWallet _privateWallet;
 
+        private const string _messagePrefix = "LACHAIN";
+
         private readonly DefaultCrypto crypto = new DefaultCrypto();
 
         public AccountServiceWeb3(IStateManager stateManager,
@@ -184,7 +186,7 @@ namespace Lachain.Core.RPC.HTTP.Web3
                 throw new ArgumentException("address should not be null");
             }
 
-            byte[]? messageBytes = message.HexToBytes();
+            byte[]? messageBytes = (_messagePrefix + message.Length + message).HexToBytes();
 
             if (_privateWallet.IsLocked())
             {
--- a/src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
+++ b/src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
@@ -188,8 +188,6 @@ namespace Lachain.Core.RPC.HTTP.Web3
                 throw new ArgumentException("address should not be null");
             }
 
-            byte[]? messageBytes = Encoding.ASCII.GetBytes(_messagePrefix + message.Length + message);
-
             if (_privateWallet.IsLocked())
             {
                 throw new Exception("wallet is locked");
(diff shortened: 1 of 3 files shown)


src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -170,10 +179,32 @@ public JArray GetAccounts()
...
+            {
+                throw new Exception("address is invalid");
+            }
+
+            var keyPair = _privateWallet.EcdsaKeyPair;
+
+            Logger.LogInformation($"Keys: {keyPair.PublicKey.GetAddress().ToHex()}");
+
+            var messageHashBytes = messageBytes.KeccakBytes();
+            var signed = crypto.SignHashed(messageHashBytes, keyPair.PrivateKey.Encode());
```
sgladkov, line 205: As I said,  don't just sign received message,  add some constant prefix to the message before signing,  for example sign smth like "LACHAIN" + messageLength + message

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -170,10 +179,32 @@ public JArray GetAccounts()
...
+
+            byte[]? messageBytes = message.HexToBytes();
+
+            if (_privateWallet.IsLocked())
+            {
+                throw new Exception("wallet is locked");
+            }
+
+            var addressUInt160 = (address!).HexToUInt160();
+            if (addressUInt160.Equals(_privateWallet.EcdsaKeyPair.PublicKey.GetAddress()))
```
tbssajal, line 195: I think it should be NotEquals

sgladkov: Good catch.)

kazunetakeda25: Hi @tbssajal Yes right :)

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -170,10 +181,32 @@ public JArray GetAccounts()
...
-            // TODO: implement message signing
-            //var addressUint160 = address.HexToUInt160();
-            return Web3DataFormatUtils.Web3Data("".HexToBytes());
-            //throw new ApplicationException("Not implemented yet");
+            if (address is null)
+            {
+                throw new ArgumentException("address should not be null");
+            }
+
+            byte[]? messageBytes = (_messagePrefix + message.Length + message).HexToBytes();
```
sgladkov, line 189: It will not work,  Prefix is not correct hex string,  so HexToBytes will fail

sgladkov (changes requested):

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -170,10 +183,32 @@ public JArray GetAccounts()
...
-            // TODO: implement message signing
-            //var addressUint160 = address.HexToUInt160();
-            return Web3DataFormatUtils.Web3Data("".HexToBytes());
-            //throw new ApplicationException("Not implemented yet");
+            if (address is null)
+            {
+                throw new ArgumentException("address should not be null");
+            }
+
+            byte[]? messageBytes = Encoding.ASCII.GetBytes(_messagePrefix + message.Length + message);
```
sgladkov, line 191: No,  it is difficult to verify this signature on the client side, Use `Encoding.ASCII.GetBytes` for conversion of the  `_messagePrefix + message.Length` and `HexToBytes` for conversion of the `message`,  after that concatenate 2 ByteArrays

sgladkov (changes requested):

sgladkov (approved):

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
```diff
@@ -12,30 +12,39 @@
...
         private readonly IStateManager _stateManager;
         private readonly ISnapshotIndexRepository _snapshotIndexer;
         private readonly IContractRegisterer _contractRegisterer;
         private readonly ISystemContractReader _systemContractReader;
         private readonly ITransactionPool _transactionPool;
+        private readonly IPrivateWallet _privateWallet;
 
-        
-
+        private readonly DefaultCrypto crypto = new DefaultCrypto();
```
tbssajal, line 37: I think it should be something like this: https://github.com/LATOKEN/lachain/blob/dev/src/Lachain.Core/Blockchain/Operations/TransactionSigner.cs#L12

tbssajal (dismissed):

src/Lachain.Core/RPC/HTTP/Web3/AccountServiceWeb3.cs
kazunetakeda25: fixed, thanks

sgladkov (dismissed):
For the future: don't add format changes to functional pull request please,  it makes review more difficult.

HMAshiqulIslam (changes requested):
Adding an extra space after "if" and "foreach" makes it inconsistent throughout the codebase. We already follow a convention and the whole codebase is based on that. We should follow the existing conventions.
=> Got it. fixed

sgladkov (approved):

HMAshiqulIslam (commented):
Can you please add some unit test / testing procedure to test this api call ?

PR #190, opened 2022-02-11
Benchmark 1000 tx time

dushyantb:


sgladkov (changes requested):

src/Lachain.Benchmark/BlockchainBenchmark.cs
```diff
@@ -181,7 +206,7 @@ private static void _Benchmark(string text, Func<int, int> action, uint tries)
                 var blockWithTxs = blocks[i];
                 // blockchainManager.PersistBlockManually(blockWithTxs.Block, blockWithTxs.Transactions);
                 return i;
-            }, (uint) blocks.Length);
+            }, (uint)blocks.Length);
 
             var elapsedTime = TimeUtils.CurrentTimeMillis() - currentTime;
```
sgladkov, line 211: Print this time too

dushyantb: this is the old benchmark function, and not touched; any ways the time is already printed

PR #193, opened 2022-02-15
Change signature algorithm for private API methods

kazunetakeda25:
Now we sign request body as a whole,  but it is not safe,  body contains spaces,  tabs,  line breaks, and other symbols that can be added or removed during transfer from client to the server. We should construct some string whose format will be predefined in hex view,  smth like this - https://api.latoken.com/doc/v2/#section/Authorization/Signature,  as an example we can use 'method' + "args,  separated with commas"

### DIFF
--- a/src/Lachain.Core/RPC/HTTP/HttpService.cs
+++ b/src/Lachain.Core/RPC/HTTP/HttpService.cs
@@ -152,7 +152,7 @@ namespace Lachain.Core.RPC.HTTP
                 var requestObj = (JObject)item;
                 if(requestObj == null) return false;
 
-                if(!_CheckAuth(requestObj, signature, timestamp))
+                if(!_CheckAuth(requestObj, context, signature, timestamp))
                 {
                     var error = new JObject
                     {
@@ -178,7 +178,7 @@ namespace Lachain.Core.RPC.HTTP
             return true;
         }
 
-        private bool _CheckAuth(JObject body, string signature, string timestamp)
+        private bool _CheckAuth(JObject body, HttpListenerContext context, string signature, string timestamp)
         {
             if(context.Request.IsLocal) return true;
 
--- a/src/Lachain.Core/RPC/HTTP/HttpService.cs
+++ b/src/Lachain.Core/RPC/HTTP/HttpService.cs
@@ -9,7 +9,6 @@ using Newtonsoft.Json.Linq;
 using Lachain.Logger;
 using Lachain.Utility.Utils;
 using Secp256k1Net;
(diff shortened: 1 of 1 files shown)


kazunetakeda25:
Hi @sgladkov 
Test repository is here: https://github.com/LATOKEN/lachain-esdca-auth-demo

src/Lachain.Core/RPC/HTTP/HttpService.cs
```diff
@@ -81,23 +81,37 @@ private bool _Handle(HttpListenerContext context)
...
+            if(timestamps != null && timestamps.Length > 0)
             {
-                sig = sigs[0];
+                timestamp = timestamps[0];
             }
 
+            // If unix timestamp diff is longer than 30 minutes, we dont handle it
+            long.TryParse(timestamp.Trim(), out long unixTimestamp);
+            TimeSpan timeSpan = DateTimeOffset.Now.Subtract(DateTimeOffset.FromUnixTimeSeconds(unixTimestamp));
+            if(timeSpan.TotalMinutes >= 30)
```
sgladkov, line 103: We should check this header for private methods only

src/Lachain.Core/RPC/HTTP/HttpService.cs
```diff
@@ -26,7 +26,7 @@ public void Start(RpcConfig rpcConfig)
                 {
                     _Worker(rpcConfig);
                 }
-                catch (Exception e)
+                catch(Exception e)
```
sgladkov, line 29: Don't add format changes to the functional pull request please

src/Lachain.Core/RPC/HTTP/HttpService.cs
```diff
@@ -169,20 +187,27 @@ private bool _Handle(HttpListenerContext context)
...
+            if(_privateMethods.Contains(body["method"]!.ToString()))
             {
-                if (string.IsNullOrEmpty(sig))
+                if(string.IsNullOrEmpty(sig))
                 {
                     return false;
                 }
 
-                var messageBytes = Encoding.UTF8.GetBytes(bodyPlainText);
+                string[] bodyParams = bodyPlainText.Split(";");
```
sgladkov, line 199: It is a wrong way to get message to sign. You have `body` here,  already parsed,  you can construct message to sign from it directly smth like `var msgToSign = body['method'] + serialize(body['params']) + timestamp`

sgladkov (changes requested):

sgladkov (approved):

PR #197, opened 2022-02-16
Benchmark tests

dushyantb:


src/Lachain.Benchmark/BlockchainBenchmark.cs
```diff
@@ -387,6 +399,51 @@ EcdsaKeyPair keyPair
...
+                {
+                    To = random.Slice(0, 20).ToUInt160(),
+                    From = keyPair.PublicKey.GetAddress(),
+                    GasPrice = (ulong)Money.Parse("0.0000001").ToWei(),
+                    GasLimit = 100000000,
+                    Nonce = (ulong)i,
+                    Value = amount
+                };
+
+                transactionPool.Add(transactionSigner.Sign(tx, keyPair), false);
```
sgladkov, line 437: Can we separate transaction creation and adding them  to pool? Tx signing can be time consuming operation and we can't optimize it,  let's don't count it. Create a txes before starting to measure the time,  place them to some array,  and after that measure the time required to add ready txes to the pool.

sgladkov (changes requested):

sgladkov (approved):

PR #206, opened 2022-02-24
[Consensus Serialization] HoneyBadger

kazunetakeda25:


src/Lachain.Consensus/HoneyBadger/HoneyBadgerId.cs
```diff
@@ -13,16 +19,16 @@ protected bool Equals(HoneyBadgerId other)
             return Era == other.Era;
         }
 
-        public bool Equals(IProtocolIdentifier other)
+        public bool Equals(IProtocolIdentifier? other)
```
sgladkov, line 22: Looks like some contradiction together with next line: in this line we tell that other can be null,  but in next line we tell that it can't be null. Why we fix it here at all?

src/Lachain.Consensus/IPublicConsensusKeySet.cs
```diff
@@ -13,5 +13,6 @@ public interface IPublicConsensusKeySet
         PublicKeySet ThresholdSignaturePublicKeySet { get; }
         IList<ECDSAPublicKey> EcdsaPublicKeySet { get; }
         public int GetValidatorIndex(ECDSAPublicKey publicKey);
+        public byte[] ToBytes();
```
sgladkov, line 16: Should we add corresponding `FromBytes` method too?

src/Lachain.Consensus/PublicConsensusKeySet.cs
```diff
@@ -18,7 +21,7 @@ public class PublicConsensusKeySet : IPublicConsensusKeySet
         public PublicConsensusKeySet(int n, int f,
             PublicKey tpkePublicKey,
             PublicKeySet thresholdSignaturePublicKeySet,
-            IEnumerable<ECDSAPublicKey> ecdsaPublicKeys
+            IEnumerable<ECDSAPublicKey> ecdsaPublicKeys/*, byte[] serializableByteObject*/
```
sgladkov, line 24: Why we have this commented code here?

sgladkov (changes requested):

src/Lachain.Consensus/PublicConsensusKeySet.cs
kazunetakeda25: I will remove comment

src/Lachain.Consensus/IPublicConsensusKeySet.cs
kazunetakeda25: I don't think so, because FromBytes is static method.

HMAshiqulIslam (changes requested):
I think current serialization won't really work. There are many state variables that need to restored. For example, 
        private readonly IRawShare?[] _shares;
        private readonly ISet<PartiallyDecryptedShare>[] _decryptedShares;
        private readonly bool[] _taken;
        private ResultStatus _requested;
        private IRawShare? _rawShare;
        private EncryptedShare? _encryptedShare;
        private ISet<IRawShare>? _result;
        private bool _takenSet;

All of the above are part of the state of this protocol.

dushyantb:
>  don't the variables which are getting set in the `constructor` are enough to complete the serialisation? or do we have to include all other variables as well?

kazunetakeda25:
> 

How can we restore private variables? Do you have a good idea?

PR #211, opened 2022-03-03
changed logic for canMissVerification

HMAshiqulIslam:


src/Lachain.Core/Blockchain/Operations/TransactionManager.cs
```diff
@@ -148,5 +154,64 @@ public ulong CalcNextTxNonce(UInt160 from)
...
+
+        private bool IsDistributeCycleRewardsAndPenaltiesTx(Block block, Transaction tx) 
+        {
+            var indexInCycle = block.Header.Index % StakingContract.CycleDuration;
+            var cycle = block.Header.Index / StakingContract.CycleDuration;
+            if (cycle > 0 && indexInCycle == StakingContract.AttendanceDetectionDuration)
+            {
+                var expectedTx = BuildSystemContractTx(ContractRegisterer.GovernanceContract, GovernanceInterface.MethodDistributeCycleRewardsAndPenalties, 
+                    UInt256Utils.ToUInt256(GovernanceContract.GetCycleByBlockNumber(block.Header.Index - 1)));
+                return expectedTx.Equals(tx);
```
sgladkov, line 166: It looks dangerous to me. Received tx can differ in some not important params like GasPrice or something like this. May be better to check inmportant fields (From,  To,  Invocation,  Value) only?

sgladkov:
Also we need detailed information in the log if we reject system transaction (exact reason and tx itself)

src/Lachain.Core/Blockchain/Operations/TransactionManager.cs
HMAshiqulIslam: Nice Catch. I totally missed this. I will push a fix.

HMAshiqulIslam:
> Also we need detailed information in the log if we reject system transaction (exact reason and tx itself)

HMAshiqulIslam:
> Also we need detailed information in the log if we reject system transaction (exact reason and tx itself)
I agree. I will write a fix. Thanks.

sgladkov (approved):

PR #212, opened 2022-03-08
Fixed logic for delegate call for transfer, create, write_log

soloway2010:


sgladkov:
I'll add a hardfork as a separate commit here

sgladkov (approved):

PR #218, opened 2022-03-14
Shrink db key

tbssajal:


sgladkov (approved):
LGTM

src/Lachain.Storage/Trie/NodeRepository.cs
```diff
@@ -44,5 +46,35 @@ public void SaveBatch(WriteBatch batch)
         {
             _rocksDbContext.SaveBatch(batch);
         }
+
+        public bool NodeIdExist(ulong id, IDbShrinkRepository _repo)
```
HMAshiqulIslam, line 50: I think these methods should not be a part of NodeRepository. Let's isolate these specific methods to some other object or (maybe in DbShrinkRepo). What's your opinion about it ?

HMAshiqulIslam (changes requested):

src/Lachain.Storage/Trie/NodeRepository.cs
tbssajal: DbShrinkRepo would be a good option.

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,118 @@
...
+using Lachain.Crypto;
+using Lachain.Storage.Trie;
+
+namespace Lachain.Storage.DbCompact
+{
+    public class DbShrinkRepository : IDbShrinkRepository
+    {
+        private IRocksDbContext _dbContext;
+        private RocksDbAtomicWrite batch;
+        private readonly ConcurrentDictionary<UInt256, byte[]> _memDbToSave = new ConcurrentDictionary<UInt256, byte[]>();
```
HMAshiqulIslam, line 13: I think maintaining two different hashMaps and underlying assumptions make it less general. Why don't we keep a Map of (hash, List<byte[]>) so that we have all the versions and we can just take the last of the list ?

HMAshiqulIslam (changes requested):

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
tbssajal: good idea!

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,119 @@
...
+            UpdateBatch();
+        }
+
+        private void UpdateBatch()
+        {
+            batch = new RocksDbAtomicWrite(_dbContext);
+            _memDb.Clear();
+        }
+
+        public void Save(byte[] key, byte[] content, bool tryCommit = true)
```
HMAshiqulIslam, line 27: Let's make this save to private.

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,119 @@
...
+            var keyHash = key.Keccak();
+            _memDb[keyHash] = content;
+            DbShrinkUtils.UpdateCounter();
+            if(tryCommit && DbShrinkUtils.CycleEnded())
+            {
+                Commit();
+            }
+        }
+
+        public void Delete(byte[] key, bool tryCommit = true)
```
HMAshiqulIslam, line 39: better to be a private method.

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,119 @@
...
+            var keyHash = key.Keccak();
+            _memDb[keyHash] = null;
+            DbShrinkUtils.UpdateCounter();
+            if(tryCommit && DbShrinkUtils.CycleEnded())
+            {
+                Commit();
+            }
+        }
+
+        public void Commit()
```
HMAshiqulIslam, line 51: better to be private.

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            batch.Commit();
+            DbShrinkUtils.ResetCounter();
+            UpdateBatch();
+        }
+
+        public byte[]? Get(byte[] key)
```
HMAshiqulIslam, line 58: better to be private

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            var keyHash = key.Keccak();
+            if (_memDb.TryGetValue(keyHash, out var content))
+            {
+                return content;
+            }
+            return _dbContext.Get(key);
+        }
+
+        public bool KeyExists(byte[] key)
```
HMAshiqulIslam, line 68: better to be private.

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            // Deleting temporary nodes
+            var prefix = EntryPrefix.NodeIdForRecentSnapshot.BuildPrefix(id);
+            Delete(prefix);
+        }
+
+        public IHashTrieNode? GetNodeById(ulong id)
+        {
+            var prefix = EntryPrefix.PersistentHashMap.BuildPrefix(id);
+            var content = Get(prefix);
```
HMAshiqulIslam, line 105: This Get would fail because the key is hashed in Get, but it should not be doing that in this case. Do you think we should remove the part of keyHashing altogether. Do we really need it ?

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (content is null) return null;
+            return NodeSerializer.FromBytes(content);
+        }
+        public void DeleteNode(ulong id , IHashTrieNode node)
+        {
+            // first delete the version by hash and then delete the node.
+            var prefix = EntryPrefix.VersionByHash.BuildPrefix(node.Hash);
+            Delete(prefix, false);
```
HMAshiqulIslam, line 113: Delete(prefix) won't work as it is hashed, but in the original db it's not hashed.

HMAshiqulIslam (changes requested):

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
tbssajal: done

src/Lachain.Storage/DbCompact/DbShrinkRepository.cs
tbssajal: done

HMAshiqulIslam (changes requested):
(1) Can we add soft compact (compactRange() of rocksDB) so that the db is compacted and we can see the reduction in DB ? 

(2) One issue: 
Suppose, we run compact hard with depth = 100. Then out of mistake, we try to compact hard with depth = 200. It will be unsucessful and throw exception which is an expected behave. But the problem is that, after that we can neither run the node nor compact hard with any smaller depth. So, we are basically stuck forever.

tbssajal:
> (1) Can we add soft compact (compactRange() of rocksDB) so that the db is compacted and we can see the reduction in DB ?
> 
> (2) One issue: Suppose, we run compact hard with depth = 100. Then out of mistake, we try to compact hard with depth = 200. It will be unsucessful and throw exception which is an expected behave. But the problem is that, after that we can neither run the node nor compact hard with any smaller depth. So, we are basically stuck forever.

(1) done
(2) nice catch. fixed.

PR #219, opened 2022-03-15
added comments and removed some unused codes

Arghya115:


src/Lachain.Storage/EntryPrefix.cs
```diff
@@ -57,7 +57,5 @@ public enum EntryPrefix : short
         SavedBatch = 0x0703,
         TotalBatch = 0x0704,
         QueueBatch = 0x0705,
-        NodesDownloadedTillNow = 0x0706,
```
sgladkov, line 60: Why do we remove this const and related code?

Arghya115: Earlier I added this code to show in the frontend how many nodes have been downloaded till now but as we don't know total number of nodes to download, we can't show progress percentage. So we didn't use this later.

tbssajal (approved):

PR #229, opened 2022-04-15
Block sync checkpoint

tbssajal:
Updated FastSync to work with BlockSynchronizer and Checkpoint


src/Lachain.Core/Blockchain/Operations/CheckpointManager.cs
```diff
@@ -0,0 +1,234 @@
...
+                Logger.LogTrace("Saved checkpoint successfully");
+                UpdateCache();
+            }
+            else Logger.LogTrace("Could not save checkpoint");
+            return saved;
+        }
+
+        private void UpdateCache()
+        {
+            _checkpointBlockId = _repository.FetchCheckpointBlockId();
```
sgladkov, line 85: It is important security requirement: checkpoints should be hardcoded in the source code,  when somebody downloads client from the repo it gets checkpoint list,  checlpoints are created by the developer team, one should change source code to chjange checkpoints

PR #230, opened 2022-04-25
Fix contract issue in testnet

sgladkov:
make hardfork_3 default deploy height for contracts

### DIFF
--- a/src/Lachain.Core/Blockchain/Hardfork/HardforkHeights.cs
+++ b/src/Lachain.Core/Blockchain/Hardfork/HardforkHeights.cs
@@ -12,6 +12,7 @@ namespace Lachain.Core.Blockchain.Hardfork
         private static ulong Hardfork_5;
         private static ulong Hardfork_6;
         private static ulong Hardfork_7;
+        private static ulong Hardfork_8;
 
         //we need this value as default deploy height
         public static ulong GetHardfork_3() 
@@ -54,6 +55,11 @@ namespace Lachain.Core.Blockchain.Hardfork
             return height >= Hardfork_7;
         }
 
+        public static bool IsHardfork_8Active(ulong height)
+        {
+            return height >= Hardfork_8;
+        }
+
         public static void SetHardforkHeights(HardforkConfig hardforkConfig)
         {
             if(alreadySet == true) 
@@ -88,6 +94,10 @@ namespace Lachain.Core.Blockchain.Hardfork
             if(hardforkConfig.Hardfork_7 is null)
                 throw new Exception("hardfork_7 is null");
             Hardfork_7 = (ulong) hardforkConfig.Hardfork_7;
+            
+            if(hardforkConfig.Hardfork_8 is null)
+                throw new Exception("hardfork_9 is null");
+            Hardfork_8 = (ulong) hardforkConfig.Hardfork_8;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
```diff
@@ -61,7 +61,11 @@ private static ulong GetDeployHeightV1(ulong currentHeight, UInt160 contract, UI
                 input, gasLimit, message).ReturnValue;
             Logger.LogInformation($"GetDeployHeight result :[{(height != null ? height.ToHex() : "null")}]");
             if (HardforkHeights.IsHardfork_6Active(currentHeight))
-                height = height ?? new byte[64];
+            {
+                height ??= HardforkHeights.IsHardfork_8Active(currentHeight) ? HardforkHeights.GetHardfork_3().ToBytes().ToArray() : new byte[64]; 
+                if(height.Length < 1)
```
tbssajal, line 66: I think there is a problem here. If height.Length == 0 then it is an empty array which is not equivalent to null I suppose. So in the previous implementation it will remain the same. But now we are assigning new byte[64] if hardfork8 is not active.

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
```diff
@@ -74,13 +78,18 @@ private static ulong GetDeployHeightV2(ulong currentHeight, UInt160 contract, UI
                     input, gasLimit, message).ReturnValue;
                 Logger.LogInformation($"GetDeployHeight result :[{(height != null ? height.ToHex() : "null")}]");
                 if (HardforkHeights.IsHardfork_6Active(currentHeight))
-                    height = height ?? new byte[64];
+                {
+                    height ??= HardforkHeights.IsHardfork_8Active(currentHeight) ? HardforkHeights.GetHardfork_3().ToBytes().ToArray() : new byte[64]; 
+                    if(height.Length < 1)
```
tbssajal, line 83: I think there is a problem here. If height.Length == 0 then it is an empty array which is not equivalent to null I suppose. So in the previous implementation it will remain the same. But now we are assigning new byte[64] if hardfork8 is not active.

tbssajal (changes requested):

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
tbssajal: I think it is better to put a single if ( HardforkHeights.IsHardfork_8Active(currentHeight) ) and handle everything there.

src/Lachain.Core/Blockchain/VM/ExternalHandler.cs
tbssajal: I think it is better to put a single if ( HardforkHeights.IsHardfork_8Active(currentHeight) ) and handle everything there.

tbssajal (approved):

src/Lachain.Core/Blockchain/Hardfork/HardforkHeights.cs
```diff
@@ -82,6 +94,10 @@ public static void SetHardforkHeights(HardforkConfig hardforkConfig)
             if(hardforkConfig.Hardfork_7 is null)
                 throw new Exception("hardfork_7 is null");
             Hardfork_7 = (ulong) hardforkConfig.Hardfork_7;
+            
+            if(hardforkConfig.Hardfork_8 is null)
+                throw new Exception("hardfork_9 is null");
```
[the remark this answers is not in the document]
sgladkov: resolved

tbssajal (approved):

PR #233, opened 2022-05-11
added validator calling method

kunalgoyal9:


src/Lachain.Core/ValidatorStatus/ValidatorStatusManager.cs
```diff
@@ -486,5 +504,35 @@ public bool IsWithdrawTriggered()
             return ValidatorAttendance.FromBytes(bytes, cycle,
                 indexInCycle >= StakingContract.AttendanceDetectionDuration);
         }
+
+        [MethodImpl(MethodImplOptions.Synchronized)]
```
tbssajal, line 508: I think we should remove '[MethodImpl(MethodImplOptions.Synchronized)]' because Run() method also has this attribute and it is always running with while loop. So this method will not be triggered from outside I think

src/Lachain.Core/ValidatorStatus/ValidatorStatusManager.cs
```diff
@@ -486,5 +504,35 @@ public bool IsWithdrawTriggered()
...
+        private void OnClientWorkerAdded(object? sender, ClientWorker clientWorker)
+        {
+            clientWorker.OnSend += OnSend;
+        }
+
+        [MethodImpl(MethodImplOptions.Synchronized)]
```
tbssajal, line 514: I think we should remove '[MethodImpl(MethodImplOptions.Synchronized)]' because Run() method also has this attribute and it is always running with while loop. So this method will not be triggered from outside I think

tbssajal (changes requested):

src/Lachain.Core/ValidatorStatus/ValidatorStatusManager.cs
kunalgoyal9: @sgladkov Can you please suggest on this?

src/Lachain.Core/ValidatorStatus/ValidatorStatusManager.cs
kunalgoyal9: @sgladkov Can you please suggest on this?

sgladkov: Agree

src/Lachain.Core/ValidatorStatus/ValidatorStatusManager.cs
sgladkov: Yes,  it is better to remove it

src/Lachain.Networking/Hub/HubConnector.cs
```diff
@@ -100,6 +100,26 @@ public void Send(byte[] publicKey, byte[] message)
             CommunicationHub.Net.Hub.Send(publicKey, CompressUtils.DeflateCompress(message).ToArray());
         }
 
+        [MethodImpl(MethodImplOptions.Synchronized)]
```
tbssajal, line 103: I think it is better to use separate lock for SendVal method, because Send method can be spammed and they have same lock, so Send method can prevent calling SendVal.

tbssajal (changes requested):

PR #235, opened 2022-05-19
Check recId value before secp256k1 calls to prevent SIGABORT

sgladkov:


src/Lachain.Crypto/DefaultCrypto.cs
```diff
@@ -91,6 +91,8 @@ public bool VerifySignatureHashed(byte[] messageHash, byte[] signature, byte[] p
 
                 var parsedSig = new byte[65];
                 var recId = (RestoreEncodedRecIdFromSignatureBuffer(signature) - 36) / 2 / TransactionUtils.ChainId (useNewChainId);
+                if (recId < 0 || recId > 3)
```
kunalgoyal9, line 94: I think if we can add `recId >= 0 && recId <= 3` here as well it is much readable, Because we are throwing exception in different format and both of them means same

src/Lachain.Crypto/DefaultCrypto.cs
```diff
@@ -152,6 +154,8 @@ public byte[] RecoverSignatureHashed(byte[] messageHash, byte[] signature, bool
                 var pk = new byte[64];
                 var encodedRecId = RestoreEncodedRecIdFromSignatureBuffer(signature);
                 var recId = (encodedRecId - 36) / 2 / TransactionUtils.ChainId(useNewChainId);
+                if (recId < 0 || recId > 3)
```
kunalgoyal9, line 157: I think if we can add `recId >= 0 && recId <= 3` here as well it is much readable, Because we are throwing exception in different format and both of them means same

test/Lachain.CryptoTest/ChainIdTest.cs
```diff
@@ -229,6 +229,8 @@ public bool VerifySignatureHashed(byte[] messageHash, byte[] signature, byte[] p
 
                 var parsedSig = new byte[65];
                 var recId = (RestoreEncodedRecIdFromSignatureBuffer(signature) - 36) / 2 / ChainId (useNewChainId);
+                if (recId < 0 || recId > 3)
```
kunalgoyal9, line 232: I think if we can add `recId >= 0 && recId <= 3` here as well it is much readable, Because we are throwing exception in different format and both of them means same

test/Lachain.CryptoTest/ChainIdTest.cs
```diff
@@ -290,6 +292,8 @@ public byte[] RecoverSignatureHashed(byte[] messageHash, byte[] signature, bool
                 var pk = new byte[64];
                 var encodedRecId = RestoreEncodedRecIdFromSignatureBuffer(signature);
                 var recId = (encodedRecId - 36) / 2 / ChainId(useNewChainId);
+                if (recId < 0 || recId > 3)
```
kunalgoyal9, line 295: I think if we can add `recId >= 0 && recId <= 3` here as well it is much readable, Because we are throwing exception in different format and both of them means same

kunalgoyal9 (approved):

tbssajal (approved):
LGTM

src/Lachain.Crypto/DefaultCrypto.cs
tbssajal, line 94: If the recId is 2 or 3, could it be an error? Because the way we decode and encode recId, I think it would be only 0 or 1.

tbssajal (commented):
Although in general recId is in range 0 to 3, but I am wondering if, in our case, recId = 2 or 3 could indicate an error.

tbssajal (approved):

PR #240, opened 2022-06-10
Validator ddos attack fix

tbssajal:
Did not test yet. Will test after the review.

### DIFF
--- a/src/Lachain.Consensus/AbstractProtocol.cs
+++ b/src/Lachain.Consensus/AbstractProtocol.cs
@@ -94,6 +94,9 @@ namespace Lachain.Consensus
                 if (Terminated) return;
                 Logger.LogTrace($"{Id}: protocol is terminated");
                 Terminated = true;
+                // We can empty the _queue because the messages will no longer be processed
+                // This will free some memory in case of spam messages
+                _queue.Clear();
                 Monitor.Pulse(_queueLock);
             }
         }
--- a/src/Lachain.Consensus/AbstractProtocol.cs
+++ b/src/Lachain.Consensus/AbstractProtocol.cs
@@ -136,6 +136,10 @@ namespace Lachain.Consensus
                 }
                 catch (Exception e)
                 {
+                    // We should investigate exceptions for each protocol and handle them carefully. Consensus depend 
+                    // on the messages from honest validators to deliver properly, no matter the delay. If a protocol
+                    // stops due to exception while handling message from malicious node, this may interrupt the communication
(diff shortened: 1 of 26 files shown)


src/Lachain.Consensus/AbstractProtocol.cs
```diff
@@ -134,7 +137,7 @@ public void Start()
                 catch (Exception e)
                 {
                     Logger.LogError($"{Id}: exception occured while processing message: {e}");
-                    Terminated = true;
+                    Terminate();
```
sgladkov, line 140: I don't sure it is correct,  we can get deadlock here because Terminate contains lock(_queueLock), so,  if we get  exception from inside other lock Terminate will never be executed really

src/Lachain.Consensus/AbstractProtocol.cs
```diff
@@ -144,8 +147,13 @@ public void ReceiveMessage(MessageEnvelope message)
...
             lock (_queueLock)
             {
-                if (Terminated){}
-            //        Logger.LogTrace($"{Id}: got message after termination"); 
+                if (Terminated)
+                {
+                    // we should return here instead of enqueueing messages
+                    // because once terminated, the messages are not being processed anymore
+                    // so the queue will just get large unnecessarily
+                    return;
```
sgladkov, line 155: The same: can we get some deadlock here due to the absence of the Monitor.Pulse?

src/Lachain.Consensus/AbstractProtocol.cs
tbssajal: understood, reverting change

src/Lachain.Consensus/AbstractProtocol.cs
tbssajal: Here adding Monitor.Pulse(_queueLock); before return will work, right?

sgladkov: I don't sure. It can unlock the thread which tries to read from queue and gets new exception

tbssajal: The queue is read only in Start() method, and it is private so not accessible from outside, also at this point the incoming message is not enqueued yet, so I think it should not be a problem.

sgladkov: Yes,  you are right

sgladkov (approved):

sgladkov (dismissed):

PR #241, opened 2022-06-16
Pool tx replacement by nonce fix

tbssajal:


src/Lachain.Core/Blockchain/Error/OperatingError.cs
```diff
@@ -27,5 +27,6 @@ public enum OperatingError : byte
         BlockGasOverflow,
         Underpriced,
         ValueOverflow,
+        BadTransaction,
```
sgladkov, line 30: I think it is better to make more narrow error type here,  smth like DuplicatedTransaction or smth like this

src/Lachain.Core/Blockchain/Pool/TransactionPool.cs
```diff
@@ -219,15 +219,32 @@ public OperatingError Add(TransactionReceipt receipt, bool notify = true)
...
+                        Logger.LogTrace(
+                            $"Transaction {receipt.Hash.ToHex()} with nonce: {receipt.Transaction.Nonce} and gasPrice: " +
+                            $"{receipt.Transaction.GasPrice} trying to replace transaction {oldTx.Hash.ToHex()} with nonce: " +
+                            $"{oldTx.Transaction.Nonce} and gasPrice: {oldTx.Transaction.GasPrice} but cannot due to lower gasPrice");
                         return OperatingError.Underpriced;
                     }
+
+                    var identicalTx = new Transaction(receipt.Transaction);
+                    identicalTx.GasPrice = oldTx.Transaction.GasPrice;
+                    if (!identicalTx.Equals(oldTx.Transaction))
```
sgladkov, line 231: I think it is better to compare From,  To Amount and Invocation explicitly,  may be inside some separate function

src/Lachain.Core/Blockchain/Error/OperatingError.cs
tbssajal: Ok

src/Lachain.Core/Blockchain/Pool/TransactionPool.cs
sgladkov: From,  To,  Value, Invocation : 4 fields

src/Lachain.Core/Blockchain/Pool/TransactionPool.cs
```diff
@@ -260,6 +275,12 @@ private bool IsBalanceValid(TransactionReceipt receipt)
             return balance.CompareTo(fee) >= 0;
         }
 
+        private bool IdenticalTransaction(Transaction oldTx, Transaction newTx)
+        {
+            return oldTx.From.Equals(newTx.From) && oldTx.To.Equals(newTx.To) &&
```
sgladkov, line 280: Value field is missed

sgladkov (approved):

PR #242, opened 2022-06-21
Fix error responce format in RPC

sgladkov:
When we throw exception in RPC method implementation response is formatted by the JsonRpc library


tbssajal (approved):
LGTM

PR #243, opened 2022-06-22
Consensus protocol crash fix

tbssajal:
Crash fix of different protocols due to exception. Did not test yet. Will test after the review.

### DIFF
--- a/src/Lachain.Consensus/HoneyBadger/HoneyBadger.cs
+++ b/src/Lachain.Consensus/HoneyBadger/HoneyBadger.cs
@@ -7,6 +7,7 @@ using Lachain.Consensus.Messages;
 using Lachain.Crypto;
 using Lachain.Crypto.TPKE;
 using Lachain.Proto;
+using Lachain.Utility.Utils;
 
 namespace Lachain.Consensus.HoneyBadger
 {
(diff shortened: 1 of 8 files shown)


src/Lachain.Consensus/HoneyBadger/HoneyBadger.cs
```diff
@@ -159,13 +160,34 @@ private ConsensusMessage CreateDecryptedMessage(PartiallyDecryptedShare share)
...
+            {
+                // Converting any random bytes to G1 is not possible
+                share = Wallet.TpkePublicKey.Decode(msg);
+                _decryptedShares[share.ShareId].Add(share);
+            }
+            catch (Exception exception)
+            {
+                var pubKey = Broadcaster.GetPublicKeyById(senderId)!.ToHex();
+                Logger.LogWarning($"Exception occured handling Decrypted message: {msg} from {senderId} ({pubKey})");
+            }
```
sgladkov, line 178: Should we re-throw this exception to the higher level?

sgladkov (approved):

src/Lachain.Consensus/HoneyBadger/HoneyBadger.cs
tbssajal: Like in RootProtocol? By the way if we throw exception then HoneyBadger will stop, right? Should we do that?

sgladkov: yes,  you are right, we should not stop the protocol in this case,  so we should not re-throw the error

PR #250, opened 2022-07-01
Private RPC methods authentication

sgladkov:


src/Lachain.Core/RPC/HTTP/HttpService.cs
```diff
@@ -179,43 +194,70 @@ private bool _Handle(HttpListenerContext context)
...
+                    var pk = new byte[64];
+                    var recId = sigBytes[64];
+                    if (recId < 0 || recId > 3)
+                        throw new Exception($"Bad signature,  invalid recId={recId}: : recId >= 0 && recId <= 3 ");
+                    if (!secp256K1.RecoverableSignatureParseCompact(parsedSig, sigBytes.Take(64).ToArray(), recId))
+                        throw new ArgumentException(nameof(signature));
+                    if (!secp256K1.Recover(pk, parsedSig, messageHash))
+                        throw new ArgumentException("Bad signature");
+                    Logger.LogTrace($"Recovered public key: {pk.ToHex()}");
+                    return pk.ToHex() == _apiKey!;
```
tbssajal, line 254: This match could be wrong in case of CloudKeygen  because the public key will be 33 bytes but pk is 64 bytes. We should generate the compressed one from pk

tbssajal (changes requested):

src/Lachain.Core/RPC/HTTP/HttpService.cs
```diff
@@ -179,43 +194,70 @@ private bool _Handle(HttpListenerContext context)
...
+                    if(!long.TryParse(timestamp.Trim(), out long unixTimestamp)) return false;
+                    TimeSpan timeSpan = DateTimeOffset.Now.Subtract(DateTimeOffset.FromUnixTimeSeconds(unixTimestamp));
+                    if(timeSpan.TotalMinutes >= 30)
+                        return false;
 
-                var messageToSign = body["method"]!.ToString() + serializedParams + timestamp;
-                var messageBytes = Encoding.UTF8.GetBytes(messageToSign);
-                var messageHash = System.Security.Cryptography.SHA256.Create().ComputeHash(messageBytes);
+                    // Serialize params (format: key1value1key2value2...)
+                    string serializedParams = SerializeParams(body["params"] as JObject);
```
tbssajal, line 230: sometime 'params' can be a list (made of JToken or JObject) instead of JObject. Will SerializeParams work here? casting body["params"] as JObject if it is a list?

tbssajal (changes requested):

src/Lachain.Core/RPC/HTTP/HttpService.cs
tbssajal: For example "la_sendRawTransactionBatch" has list of strings in params

sgladkov: Fixed

kunalgoyal9 (approved):

tbssajal (approved):
LGTM

PR #254, opened 2022-07-14
Fix for eth_estimateGas RPC method

sgladkov:
Make a transfer before contract call if required

### DIFF
--- a/src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
+++ b/src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
@@ -475,6 +475,19 @@ namespace Lachain.Core.RPC.HTTP.Web3
                 InvocationResult invRes = _stateManager.SafeContext(() =>
                 {
                     var snapshot = _stateManager.NewSnapshot();
+					if (!tx.Value.IsZero())
+					{
+                        var transferContext = new InvocationContext(source, snapshot, new TransactionReceipt
+                        {
+                            Block = snapshot.Blocks.GetTotalBlockHeight(),
+                            Transaction = tx
+                        });
+                    
+                        var localInvocation = ContractEncoder.Encode("transfer(address,uint256)", destination, tx.Value);
+                        var transferResult =
+                            ContractInvoker.Invoke(ContractRegisterer.LatokenContract, transferContext, localInvocation, GasMetering.DefaultBlockGasLimit);
+                        gasUsed += transferResult.GasUsed;
+                    }
                     var res = VirtualMachine.InvokeWasmContract(
                         contract,
                         new InvocationContext(source, snapshot, new TransactionReceipt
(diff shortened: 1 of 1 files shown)


tbssajal (approved):

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
```diff
@@ -475,6 +475,19 @@ public string Call(JObject opts, string? blockId)
...
+                        var transferContext = new InvocationContext(source, snapshot, new TransactionReceipt
+                        {
+                            Block = snapshot.Blocks.GetTotalBlockHeight(),
+                            Transaction = tx
+                        });
+                    
+                        var localInvocation = ContractEncoder.Encode("transfer(address,uint256)", destination, tx.Value);
+                        var transferResult =
+                            ContractInvoker.Invoke(ContractRegisterer.LatokenContract, transferContext, localInvocation, GasMetering.DefaultBlockGasLimit);
+                        gasUsed += transferResult.GasUsed;
```
kunalgoyal9, line 489: @sgladkov Should we rollback stateManager here using `_stateManager.Rollback();`?

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
```diff
@@ -496,6 +509,19 @@ public string Call(JObject opts, string? blockId)
...
+                    var transferContext = new InvocationContext(source, snapshot, new TransactionReceipt
+                    {
+                        Block = snapshot.Blocks.GetTotalBlockHeight(),
+                        Transaction = tx
+                    });
+                
+                    var localInvocation = ContractEncoder.Encode("transfer(address,uint256)", destination, tx.Value);
+                    var transferResult =
+                        ContractInvoker.Invoke(ContractRegisterer.LatokenContract, transferContext, localInvocation, GasMetering.DefaultBlockGasLimit);
+                    gasUsed += transferResult.GasUsed;
```
kunalgoyal9, line 523: @sgladkov Same here
Should we rollback stateManager here using _stateManager.Rollback();?

kunalgoyal9 (changes requested):

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
sgladkov: No, we make rollback after the script execution. The goal of this fix is to make transferred funds available to the contract during its execution emulation

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
sgladkov: the same

PR #256, opened 2022-07-21
fixed address, public key, balance format

tbssajal:
Helps to test

### DIFF
--- a/src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
+++ b/src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
@@ -452,7 +452,7 @@ namespace Lachain.Core.RPC.HTTP.Web3
 
             var balance = _stateManager.CurrentSnapshot.Balances.GetBalance(addressUint160);
 
-            var stake = _systemContractReader.GetStake(addressUint160).ToMoney().ToWei();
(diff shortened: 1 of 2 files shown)


src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
```diff
@@ -452,7 +452,7 @@ public JObject GetValidatorInfo(string publicKeyStr)
 
             var balance = _stateManager.CurrentSnapshot.Balances.GetBalance(addressUint160);
 
-            var stake = _systemContractReader.GetStake(addressUint160).ToMoney().ToWei();
+            var stake = _systemContractReader.GetStake(addressUint160).ToMoney();
```
sgladkov, line 455: Why is it better to return it in LA not wei?

tbssajal: The balance is also returned in LA, so it is easy to compare, I guess. And in fe_account api, it is returned in LA. So I thought they would be similar.

sgladkov: ok

sgladkov (approved):

kunalgoyal9 (approved):

PR #257, opened 2022-07-22
Api fix and update

tbssajal:
Seems like using IValidatorManager in BlockchainServiceWeb3 gives some errors in Pull Request Build. Using ISnapshotIndexRepository builds properly

### DIFF
--- a/src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
+++ b/src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
@@ -1,26 +1,27 @@
 using System;
 using System.Linq;
 using AustinHarris.JsonRpc;
+using Lachain.Core.Blockchain.Hardfork;
 using Lachain.Core.Blockchain.Interface;
 using Lachain.Core.Blockchain.Pool;
+using Lachain.Core.Blockchain.SystemContracts;
+using Lachain.Core.Blockchain.Validators;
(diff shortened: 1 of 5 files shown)


src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
```diff
@@ -712,6 +716,21 @@ private JArray GetConsensusState()
             return protocols;
         }
 
+        [JsonRpcMethod("la_getValidatorsAfterBlock")]
+        public JArray GetValidatorsAfterBlock(string blockTag)
+        {
+            if (blockTag == "pending")
```
kunalgoyal9, line 722: Shouldn't we handle cases like blockTag doens't belong to {"pending", "latest"}?

sgladkov (approved):

kunalgoyal9 (approved):

src/Lachain.Core/RPC/HTTP/Web3/BlockchainServiceWeb3.cs
tbssajal: Sorry did not understand the question

sgladkov: We are checking "pending" value only because all other values are processed in  `GetBlockNumberByTag` call @kunalgoyal9

sgladkov (approved):

PR #262, opened 2022-08-09
Validator attendance fix

tbssajal:
tested, works for me


src/Lachain.Consensus/RootProtocol/RootProtocol.cs
```diff
@@ -277,7 +280,10 @@ private void CheckSignatures()
 
             try
             {
+                if (_validatorAttendance is null)
+                    throw new Exception("validator attendance cannot be null");
```
sgladkov, line 284: I don't think we should stop block producing due to the absent attendance,  just save its state if it is available. it is ok not to send attendance at all

sgladkov (changes requested):

src/Lachain.Consensus/RootProtocol/RootProtocol.cs
tbssajal: thanks, fixed

sgladkov (approved):

PR #263, opened 2022-08-12
Prioritizing consensus message

tbssajal:
Tested

### DIFF
--- a/src/Lachain.Networking/Hub/ClientWorker.cs
+++ b/src/Lachain.Networking/Hub/ClientWorker.cs
@@ -36,8 +36,8 @@ namespace Lachain.Networking.Hub
         private int _eraMsgCounter;
         private readonly object _messageReceived = new object();
 
-        private readonly Queue<NetworkMessage> _messageQueue = new Queue<NetworkMessage>();
-        private readonly Queue<NetworkMessage> _consensusMessages = new Queue<NetworkMessage>();
+        private readonly Queue<NetworkMessage> _nonPriorityMessageQueue = new Queue<NetworkMessage>();
+        private readonly Queue<NetworkMessage> _priorityMessageQueue = new Queue<NetworkMessage>();
 
         public ClientWorker(ECDSAPublicKey peerPublicKey, IMessageFactory messageFactory, HubConnector hubConnector)
         {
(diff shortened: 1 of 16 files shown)


src/Lachain.Networking/Hub/HubConnector.cs
```diff
@@ -100,13 +106,53 @@ public void Send(byte[] publicKey, byte[] message)
...
+                        while (_messageQueue.Count > 0)
+                        {
+                            byte[] publicKey;
+                            byte[] message;
+                            (publicKey, message) = _messageQueue.Dequeue();
+                            Send(publicKey, message);
+                        }
+                    }       
+
+                    Thread.Sleep(waitForSomeTime);
```
sgladkov, line 139: Why we need thios sleep here after waiting monitor at the cycle beginnig?

tbssajal: I am really not sure about the design here. All clientworker use this to send non priority message and lock this so I added sleep. priority messages go directly to Send method. In case of spam request (like block request) or pool sync (we have pool sync with one tx at a time, maybe  we can send several txes together?) this method can be locked for long time. So I added sleep. But I am open to suggestions.

tbssajal: I have removed this way of sending message for now, will test and see what happpes.

src/Lachain.Networking/Hub/ClientWorker.cs
```diff
@@ -61,74 +62,95 @@ public void Start()
...
             }
         }
 
         public void Dispose()
         {
-            lock (_messageQueue) _messageQueue.Clear();
+            lock (_messageQueue)
+            {
+                while (_messageQueue.Count > 0)
+                    _messageQueue.DeleteMin();
```
sgladkov, line 150: Why not Clear here? Why do you delete it one by one?

src/Lachain.Networking/Hub/ClientWorker.cs
```diff
@@ -61,74 +62,95 @@ public void Start()
...
+
+                        _hubConnector.Send(PeerPublicKey, megaBatchBytes);
+                        _eraMsgCounter += 1;
                     }
-
-                    _hubConnector.Send(PeerPublicKey, megaBatchBytes);
-                    _eraMsgCounter += 1;
+                    
+                    var toSleep = Math.Clamp(250 - (long) (TimeUtils.CurrentTimeMillis() - now), 1, 1000);
+                    Thread.Sleep(TimeSpan.FromMilliseconds(toSleep));
```
sgladkov, line 136: Are we need this sleep? We have a Wait several lines ago

tbssajal: I am not sure, it was already present so did not remove. Should I remove it and test?

sgladkov: ok,  let's leave it for now,  will remove it later

sgladkov (approved):

PR #264, opened 2022-08-12
Transaction verification optimization

tbssajal:
Tested


test/Lachain.CoreTest/IntegrationTests/BlocksTest.cs
```diff
@@ -260,6 +264,116 @@ public void Test_Wrong_Block_Emulation()
             Assert.AreEqual(OperatingError.InvalidMultisig, result);
         }
 
+        [Test]
+        public void Test_TxVerifierPerformanceCompare()
```
sgladkov, line 268: Add Ignore attribute to this test for now please,  we should decrease the number of long unit tests and move them to the separate test project because we already near the timeout of the CI process

sgladkov (dismissed):
LGTM,  but disable long test for now (using Ignore attribute)

test/Lachain.CoreTest/IntegrationTests/BlocksTest.cs
tbssajal: done!

sgladkov (approved):

PR #266, opened 2022-08-16
Partially decrypted shares from peers validation

sgladkov:


sgladkov:
Current state: local network on this branch is working,  sometimes it breaks during the validator change,  TPKE test is failing

src/Lachain.Crypto/TPKE/TrustedKeyGen.cs
```diff
@@ -27,5 +27,11 @@ public PrivateKey GetPrivKey(int i)
         {
             return new PrivateKey(MclBls12381.EvaluatePolynomial(_coeffs, Fr.FromInt(i + 1)), i);
         }
+        
+        public PublicKey GetVerificationPubKey(int i)
+        {
+            return new PublicKey(G1.Generator * MclBls12381.EvaluatePolynomial(_coeffs, Fr.FromInt(i + 1)), i);
```
tbssajal, line 33: I think the 2nd input in PublicKey(_ , i) should be _degree, I think it is faulty node count

tbssajal (changes requested):

src/Lachain.Crypto/TPKE/TrustedKeyGen.cs
sgladkov: yes,  you are right, copy-paste error.)

src/Lachain.Consensus/HoneyBadger/HoneyBadger.cs
```diff
@@ -136,6 +138,19 @@ private void HandleCommonSubset(ProtocolResult<CommonSubsetId, ISet<EncryptedSha
...
                 _receivedShares[share.Id] = share;
+                if (_decryptedShares[share.Id].Count > 0) // if we have any partially decrypted shares for this share - verify them
+                {
+                    if (!_skipDecryptedShareValidation)
+                    {
+                        if (Wallet.GetTpkeVerificationKey(share.Id) is null)
+                            _decryptedShares[share.Id].Clear();
+                        else
+                            _decryptedShares[share.Id] = _decryptedShares[share.Id]
+                                .Where(ps => Wallet.GetTpkeVerificationKey(share.Id)!.VerifyShare(share, ps))
```
tbssajal, line 149: I think it should be Wallet.GetTpkeVerificationKey(ps.DecryptorId) to get the public key of decryptor

tbssajal (changes requested):

src/Lachain.Consensus/HoneyBadger/HoneyBadger.cs
sgladkov: You are right,  thank you

src/Lachain.Core/Config/ConfigManager.cs
```diff
@@ -385,6 +387,30 @@ private void _UpdateConfigToV13()
...
+        {
+            var network = GetConfig<NetworkConfig>("network") ??
+                          throw new ApplicationException("No network section in config");
+            var hardforks = GetConfig<HardforkConfig>("hardfork") ??
+                            throw new ApplicationException("No hardfork section in config");
+            hardforks.Hardfork_12 ??= network.NetworkName switch
+            {
+                "mainnet" => 0,
+                "testnet" => 0,
+                "devnet" => 0,
```
tbssajal, line 401: we should use positive value for devnet and local network. otherwise we never skip verification and it halts consensus for F > 1

tbssajal (changes requested):

src/Lachain.Core/Config/ConfigManager.cs
tbssajal: it is because _degree number of coefficients for tpke key generation I guess.

sgladkov: Let's set it non-zero for now,  I'll add a task to investigate it later

src/Lachain.Utility/ConsensusState.cs
```diff
@@ -27,11 +36,26 @@ public static ConsensusState FromBytes(ReadOnlySpan<byte> bytes)
         {
             var decoded = (RLPCollection) RLP.Decode(bytes.ToArray());
             var tpkePubKey = decoded[0].RLPData;
-            var credentials = decoded.Skip(1)
+            if (decoded[1].RLPData.Length == 1) // new data format with verification keys
```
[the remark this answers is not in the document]
sgladkov: We can't have so many validators,  that's why I have decided to use byte here to simplify parsing

src/Lachain.Utility/ConsensusState.cs
```diff
@@ -1,24 +1,33 @@
...
+        public byte[][] TpkeVerificationKeys { get; }
         public ValidatorCredentials[] Validators { get; }
 
-        public byte[] ToBytes()
+        public byte[] ToBytes(bool useNewFormat)
         {
             var a = new List<byte[]> {TpkePublicKey};
+            if (useNewFormat)
+            {
+                a.Add(new byte[] {(byte) TpkeVerificationKeys.Length});
```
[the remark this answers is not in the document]
sgladkov: We can't have so many validators, that's why I have decided to use byte here to simplify parsing

src/Lachain.Core/Blockchain/SystemContracts/GovernanceContract.cs
```diff
@@ -312,9 +324,9 @@ public ExecutionStatus DistributeCycleRewardsAndPenalties(UInt256 cycle, SystemC
             _lastSuccessfulKeygenBlock.Set(new BigInteger(frame.InvocationContext.Receipt.Block).ToUInt256().ToBytes());
             SetPlayersCount(players);
             SetTSKeys(tsKeys);
-            SetTpkeKey(tpkePublicKey);
+            SetTpkeKey(tpkePublicKey, tpkeVerificationKeys);
 
-            Emit(GovernanceInterface.EventKeygenConfirm, tpkePublicKey, thresholdSignaturePublicKeys);
+            Emit(GovernanceInterface.EventKeygenConfirm, tpkePublicKey, tpkeVerificationKeys, thresholdSignaturePublicKeys);
```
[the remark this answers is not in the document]
sgladkov: I haven't changed event signature here really,  so it is better not to send this event at all,  we can add it in the future if required. We need some logic for smooth update of the system contracts,  let's keep it for the future. I'm going to remove verification keys for this event now.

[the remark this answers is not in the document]
tbssajal: We are emitting so event is added I think. Actually I tested hardfork yesterday and for new node state hash does not match and it stops at the middle of the cycle

src/Lachain.Core/Vault/KeyGenManager.cs
```diff
@@ -273,7 +273,7 @@ private void BlockManagerOnSystemContractInvoked(object _, InvocationContext con
                 var tpkePublicKey =
                     PublicKey.FromBytes(args[1] as byte[] ?? throw new Exception("Failed to get tpkePublicKey"));
                 var tsKeys = new PublicKeySet(
-                    (args[2] as byte[][] ?? throw new Exception()).Select(x =>
+                    (args[3] as byte[][] ?? throw new Exception()).Select(x =>
```
tbssajal, line 276: I think we should keep tskeys in args[2] and tpke keys at the last. New code tries to find tsKeys in args[3] and get error.

tbssajal (changes requested):

src/Lachain.Core/Blockchain/SystemContracts/GovernanceContract.cs
```diff
@@ -269,11 +281,11 @@ public ExecutionStatus DistributeCycleRewardsAndPenalties(UInt256 cycle, SystemC
         }
 
         [ContractMethod(GovernanceInterface.MethodKeygenConfirm)]
-        public ExecutionStatus KeyGenConfirm(UInt256 cycle, byte[] tpkePublicKey, byte[][] thresholdSignaturePublicKeys,
+        public ExecutionStatus KeyGenConfirm(UInt256 cycle, byte[] tpkePublicKey, byte[][] tpkeVerificationKeys, byte[][] thresholdSignaturePublicKeys,
```
tbssajal, line 284: node without hardfork 12 will send 3 parameters with tx. node with hardfork 12 will expect 4 parameters. This could cause error. Maybe we can set a default value for tpkeVerificationKeys ?

tbssajal (changes requested):

tbssajal (approved):

PR #273, opened 2022-09-06
Update dependencies

sgladkov:
Newsoft,Json 12.0.3 => 13.0.1
Nethereum 3.8.0 => 4.0.0


src/Lachain.Crypto/TransactionUtils.cs
```diff
@@ -26,36 +26,67 @@ public static int ChainId(bool useNewId)
...
                 t.To.ToBytes(), // this may be empty, same as passing null
                 t.Value.ToBytes(false,  true),
                 t.Invocation.ToArray(),
-                new BigInteger(ChainId(useNewId)).ToByteArray().Reverse().ToArray().TrimLeadingZeros(),
+                new BigInteger(ChainId(useNewId)).ToByteArray().Reverse().ToArray().TrimLeadingZeros(), null, null},
                 sig.IsEmpty ? Array.Empty<byte>() : sig.Slice(0, 32).ToArray().TrimLeadingZeros(),
                 sig.IsEmpty ? Array.Empty<byte>() : sig.Slice(32, 32).ToArray().TrimLeadingZeros(),
-                sig.IsEmpty ? Array.Empty<byte>() : sig.Slice(64, sig.Length - 64).ToArray().TrimLeadingZeros()
+                sig.IsEmpty ? Array.Empty<byte>() : sig.Slice(64, sig.Length - 64).ToArray().TrimLeadingZeros(), 
+                6
```
tbssajal, line 46: Not sure why using 6 here, is it the size of input byte[][] array? I see 9 elements including 2 null input. So should it be 9 instead of 6? Also maybe we can use a constant instead of number because it is used in other methods also.

src/Lachain.Crypto/TransactionUtils.cs
```diff
@@ -26,36 +26,67 @@ public static int ChainId(bool useNewId)
...
             );
-            return ethTx;
+            return rlpSigner;
         }
         public static IEnumerable<byte> Rlp(this Transaction t, bool useNewId)
         {
             var ethTx = t.GetEthTx(null, useNewId);
             return ethTx.GetRLPEncodedRaw();
```
tbssajal, line 53: should we use LAEncodeSigned here also?

tbssajal (changes requested):

src/Lachain.Core/RPC/HTTP/FrontEnd/FrontEndService.cs
```diff
@@ -338,7 +338,7 @@ public bool VerifySign(string message, string sign, bool useNewChainId)
         [JsonRpcMethod("fe_verifyRawTransaction")]
         public JObject VerifyRawTransaction(string rawTx, string externalTxHash, bool useNewChainId)
         {
-            var ethTx = new TransactionChainId(rawTx.HexToBytes());
+            var ethTx = new LegacyTransactionChainId(rawTx.HexToBytes());
```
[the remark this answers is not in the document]
sgladkov: We are using the class from other lib and they has renamed this class. I don't think it is a good idea to wrap it really,  It is better to remove Nethereum dependency at all in the future

src/Lachain.Crypto/TransactionUtils.cs
sgladkov: Nethereum doesn't use this param really in this case,  it uses constant value,  but we can set any number here really,

sgladkov: I'll add a comment here

tbssajal (approved):
Did not test with if hardfork is needed

PR #277, opened 2022-09-08
Fix consensus message handling

tbssajal:
tested


src/Lachain.Core/Consensus/ConsensusManager.cs
```diff
@@ -145,8 +146,8 @@ private bool IsEraNearFuture(long era, long currentEra)
         {
             // allow message if consensus has not started yet.
             if (currentEra == -1) return true;
-            long allowedEra = 5;
             if (era < currentEra) return false;
+            long allowedEra = 2 * (long) StakingContract.CycleDuration;
```
sgladkov, line 150: Can we be sure this value is initialized already here?

sgladkov (approved):

sgladkov (approved):

PR #278, opened 2022-09-14
Optimize pool sync request

tbssajal:


src/Lachain.Core/Blockchain/Pool/TransactionPool.cs
```diff
@@ -71,6 +75,50 @@ ITransactionHashTrackerByNonce transactionHashTracker
...
+                    var txes = new List<TransactionReceipt>();
+                    while (_transactionToSync.Count > 0)
+                    {
+                        var tx = _transactionToSync.Dequeue();
+                        txes.Add(tx);
+                    }
+                    TransactionAdded?.Invoke(this, txes);
+                }
+                // wait for several txes to appear, so we send several txes together in pool-sync
+                Thread.Sleep(waitingTime);
```
sgladkov, line 102: We should resend txes as fast as we get it,  no waiting . If we have only one tx we should send this tx only

tbssajal: Then how you suggest we send txes in batch? because this way always one tx will be sent at a time.

PR #282, opened 2022-10-04
Check account balance in estimateGas call

sgladkov:


Anachor (approved):

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
```diff
@@ -483,7 +483,7 @@ public string Call(JObject opts, string? blockId)
         
                 return transferInvRes.Status == ExecutionStatus.Ok
                     ? (gasUsed + transferInvRes.GasUsed).ToHex()
-                    : Web3DataFormatUtils.Web3Number(gasUsed);
+                    : throw new Exception("Error in funds transfer");
```
tbssajal, line 486: Maybe we can use rpc exception as we know the reason of error. It could be cleaner.

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
```diff
@@ -502,6 +502,8 @@ public string Call(JObject opts, string? blockId)
                         var localInvocation = ContractEncoder.Encode("transfer(address,uint256)", destination, tx.Value);
                         var transferResult =
                             ContractInvoker.Invoke(ContractRegisterer.LatokenContract, transferContext, localInvocation, GasMetering.DefaultBlockGasLimit);
+                        if (transferResult.Status != ExecutionStatus.Ok)
```
tbssajal, line 505: should we rollback the snapshot before returing?

sgladkov: It wiil be rolled back automatically in StateManager.Release()

src/Lachain.Core/RPC/HTTP/Web3/TransactionServiceWeb3.cs
sgladkov: ok,  I'll add it

PR #286, opened 2022-10-05
Improved transaction finality time

Anachor:
Improved the txn finality time by decrease block time to 4 sec. To preserve tps, transactions per block was reduced to 800.

- Added `targetTransactionPerBlock` field in `BlockchainConfig`
- Implemented `_UpdateConfigToV16()`
- Used `targetTransactionPerBlock` from config instead of hardcoded value in `BlockProducer.cs`

### DIFF
--- a/src/Lachain.Core/Config/ConfigManager.cs
+++ b/src/Lachain.Core/Config/ConfigManager.cs
@@ -6,12 +6,14 @@ using Lachain.Core.CLI;
 using Lachain.Networking;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Lachain.Core.Blockchain;
+
 
 namespace Lachain.Core.Config
 {
     public class ConfigManager : IConfigManager
     {
-        private const ulong _CurrentVersion = 15;
+        private const ulong _CurrentVersion = 16;
         private IDictionary<string, object> _config;
         public string ConfigPath { get; }
         public RunOptions CommandLineOptions { get; }
@@ -69,6 +71,8 @@ namespace Lachain.Core.Config
                 _UpdateConfigToV14();
             if (version < 15)
                 _UpdateConfigToV15();
+            if (version < 16)
+                _UpdateConfigToV16();
         }
 
         // version 2 of config should contain hardfork section and height for first hardfork,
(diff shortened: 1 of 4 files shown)


src/Lachain.Core/Config/ConfigManager.cs
```diff
@@ -437,6 +441,29 @@ private void _UpdateConfigToV15()
...
+        // and new blockchain.TargetTransactionsPerBlock (800) 
+        private void _UpdateConfigToV16()
+        {   
+            const int newTargetBlockTime = 4000; //ms
+            const int newTargetTransactionsPerBlock = 800; 
+
+            var blockchain = GetConfig<BlockchainConfig>("blockchain") ??
+                            throw new ApplicationException("No blockchain section in config");
+
+            blockchain.TargetBlockTime = newTargetBlockTime;
```
sgladkov, line 454: Let's replace this value only if it was 5000 before

src/Lachain.Core/Consensus/BlockProducer.cs
```diff
@@ -64,14 +66,15 @@ ITransactionBuilder transactionBuilder
             _blockManager = blockManager;
             _stateManager = stateManager;
             _transactionBuilder = transactionBuilder;
+            txPerBlock = Convert.ToInt32(configManager.GetConfig<BlockchainConfig>("blockchain")?.TargetTransactionsPerBlock ?? 1_000);
```
sgladkov, line 69: I think it is better to declare TargetTransactionsPerBlock as int in BlockchainConfig instead of this conversion

sgladkov (commented):
I have 2 small comments,  but it looks good to me

src/Lachain.Core/Config/ConfigManager.cs
Anachor: can it be anything other than 5000 before? if so, when? and why should we be not changing it in that case?

src/Lachain.Core/Consensus/BlockProducer.cs
Anachor: okay. updating.

Anachor:
Updated code addressing @sgladkov 's comments.

src/Lachain.Core/Config/ConfigManager.cs
sgladkov: it could be different from 5000 if someone has changed it manually in ts own config. In this case let him has its custom value.

sgladkov (approved):

src/Lachain.Core/Config/ConfigManager.cs
```diff
@@ -6,12 +6,14 @@
 using Lachain.Networking;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Lachain.Core.Blockchain;
```
tbssajal, line 9: try to use them in alphabetical order please

src/Lachain.Core/Consensus/BlockProducer.cs
```diff
@@ -18,7 +18,8 @@
 using Lachain.Proto;
 using Lachain.Storage.State;
 using Lachain.Utility.Utils;
-
+using Lachain.Core.Config;
```
tbssajal, line 21: try to use them in alphabetical order please

Anachor: Updated in the latest commit.

tbssajal (approved):

PR #287, opened 2022-10-07
Overflow and nonce bug fix

tbssajal:


test/Lachain.CoreTest/IntegrationTests/BlocksTest.cs
```diff
@@ -223,6 +229,58 @@ public void Test_Block_With_Txs_Execution()
             }
             ExecuteTxesInSeveralBlocks(randomReceipts);
         }
+        
+        [Test]
+        public void Test_Test()
```
sgladkov, line 234: I think it is better to give this test more explanatory name.)

sgladkov (changes requested):

test/Lachain.CoreTest/IntegrationTests/BlocksTest.cs
tbssajal: Sorry, forgot to update

sgladkov (approved):

PR #288, opened 2022-10-11
Implement backup mechanism for config

Anachor:
- When loading, first tried to load from `ConfigPath`. If failed, tried from `ConfigBackupPath`. If failed again, throw exception.
- When saving config, saved both to `ConfigPath` and `ConfigBackupPath`
- Added a call to `SaveConfig` after loading to sync config and backup


Anachor:
Questions:
- Should `ConfigBackupPath` be read from options instead of appending `.bak`?
- Should I catch more specific exceptions?

src/Lachain.Core/Config/ConfigManager.cs
```diff
@@ -440,6 +466,7 @@ private void _UpdateConfigToV15()
         private void _SaveCurrentConfig()
         {
             File.WriteAllText(ConfigPath, JsonConvert.SerializeObject(_config, Formatting.Indented));
+            File.WriteAllText(ConfigBackupPath, JsonConvert.SerializeObject(_config, Formatting.Indented));
```
tbssajal, line 469: I think backup should be saved in e separate function after updating is done. Because the case was config was corrupted during updating the config. So, in current implementation, backup could also be corrupted, right?

tbssajal (changes requested):

src/Lachain.Core/Config/ConfigManager.cs
sgladkov: Steps during config update:
1. Copy old config to config backup
2. Write new data to config
3. Delete backup
I don't think we need separate function for this,  but step order is important

src/Lachain.Core/Config/ConfigManager.cs
```diff
@@ -14,15 +15,40 @@ public class ConfigManager : IConfigManager
...
+            
+            try {
+                var configLoader = new LocalConfigLoader(ConfigPath);
+                _config = new Dictionary<string, object>(configLoader.LoadConfig());
+                Logger.LogInformation($"Loaded config from {ConfigPath}");
+            }
+            catch (Exception e) {
+                try {
+                    Logger.LogWarning($"Could not load config from {ConfigPath}. Trying from backup ({ConfigBackupPath})");
+                    var configLoader = new LocalConfigLoader(ConfigBackupPath);
```
sgladkov, line 38: In this case you need to copy backup file to the config file

sgladkov (changes requested):

src/Lachain.Core/Config/ConfigManager.cs
Anachor: This will be done in the following call to _SaveCurrentConfig()

src/Lachain.Core/Config/ConfigManager.cs
Anachor: okay, I did it slightly different way. I'm always keeping a backup file. This way, both the Config file and the backup file cannot be corrupted at the same time. In case there is a problem updating the config file, the backup file will still be intact. 

I think this solution will handle your case and also handle any case where the config file is somehow corrupted.

src/Lachain.Core/Config/ConfigManager.cs
Anachor: The config will be saved to the config file in the following call to `_SaveCurrentConfig`

sgladkov: It is not guaranteed really (for example,  exceptions on the intermediate steps). It is better to copy this file explicitly

sgladkov:
One more thing. It was not in original task conditions,  but save config in constructor only if it was really changed (add modify flag and set it if you change any data)

src/Lachain.Core/Config/ConfigManager.cs
Anachor: done in the latest commit.

Anachor:
> One more thing. It was not in original task conditions, but save config in constructor only if it was really changed (add modify flag and set it if you change any data)

I have now called save config only if something is changed.

sgladkov (changes requested):
Tests are failing due to the fail to find config.json.bak: https://app.travis-ci.com/github/LATOKEN/lachain/builds/256617012 Fix it please,  this class should not require bak file existence

Anachor:
It seems that the tests are failing because of the new test I added. But it should not happen if tests are completely independent. In my test, I try deleting config file and backup to see if everything works correctly. This deletion is causing problem in other tests.

sgladkov:
I think we don't need such test in unit test really. It is enough to test it manually. Unit tests share the same setup,  it is dangerous to delete common config

Anachor:
Removed the offending lines in the test. All checks pass now.

PR #293, opened 2022-10-17
Separate validator stream

tbssajal:


sgladkov:
Change network version in https://github.com/LATOKEN/lachain/blob/f218953d783e6efc870e606f4e30b4224310a511/src/Lachain.Core/Network/NetworkManager.cs#L10, because old clients will not be compatible with this version

PR #294, opened 2022-10-19
asking for blocks to any peer that has height greater than me

tbssajal:
tested

### DIFF
--- a/src/Lachain.Core/Network/BlockSynchronizer.cs
+++ b/src/Lachain.Core/Network/BlockSynchronizer.cs
@@ -39,6 +39,7 @@ namespace Lachain.Core.Network
 
         private readonly object _peerHasTransactions = new object();
         private readonly object _peerHasBlocks = new object();
+        private readonly object _peerHeightUpdate = new object();
         private readonly object _blocksLock = new object();
         private readonly object _txLock = new object();
         private LogLevel _logLevelForSync = LogLevel.Trace;
@@ -148,6 +149,9 @@ namespace Lachain.Core.Network
                     Logger.LogTrace($"Skipped block {block.Header.Index} from peer {publicKey.ToHex()}: invalid multisig with error: {error}");
                     return false;
                 }
+                // let it know that we received a valid response
+                lock (_peerHasBlocks)
+                    Monitor.PulseAll(_peerHasBlocks);
                 // This is to tell consensus manager to terminate current era, since we trust given multisig
                 OnSignedBlockReceived?.Invoke(this, block.Header.Index);
 
@@ -173,8 +177,6 @@ namespace Lachain.Core.Network
                     return false;
                 }
 
-                lock (_peerHasBlocks)
-                    Monitor.PulseAll(_peerHasBlocks);
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


src/Lachain.Core/Network/BlockSynchronizer.cs
```diff
@@ -296,32 +298,31 @@ private void BlockSyncWorker()
...
                         _networkManager.SendTo(
-                            peer, _networkManager.MessageFactory.SyncBlocksRequest(leftBound, rightBound),
+                            peer.Key, _networkManager.MessageFactory.SyncBlocksRequest(leftBound, rightBound),
                             NetworkMessagePriority.PeerSyncMessage
                         );
                     }
 
-                    var waitStart = TimeUtils.CurrentTimeMillis();
-                    while (true)
+                    var waitForBlockExecution = 4000;
```
sgladkov, line 317: Why do we use 4 secs here?

src/Lachain.Core/Network/BlockSynchronizer.cs
```diff
@@ -296,32 +298,31 @@ private void BlockSyncWorker()
...
-                        {
-                            Monitor.Wait(_peerHasBlocks, TimeSpan.FromMilliseconds(1_000));
-                        }
-
-                        if (TimeUtils.CurrentTimeMillis() - waitStart > 5_000) break;
+                        gotResponse = Monitor.Wait(_peerHasBlocks, TimeSpan.FromMilliseconds(1_000));
+                    }
+                    if (gotResponse)
+                    {
+                        Thread.Sleep(waitForBlockExecution);
```
sgladkov, line 325: Are we really need to use Sleep instead  of waiting for event?

src/Lachain.Core/Network/BlockSynchronizer.cs
tbssajal: it was hardcoded 5 secs, but assuming we wait 1 sec (max) for reply, I reduced to 4 secs. should I change it to 5 secs?

src/Lachain.Core/Network/BlockSynchronizer.cs
tbssajal: we wait 1 sec for reply, if we get no reply, we don't sleep, send another request, but if we get reply, then I think we should wait for some time for the incoming blocks to finish execution. I can try to handle it in such a way that we wait until execution of receiving blocks is finished and not use hardcoded value. Should I do it?

src/Lachain.Core/Network/BlockSynchronizer.cs
sgladkov: Our block times are set in config,  I think in this case we should read it from it here (now block time is 4 secs,  for example)

src/Lachain.Core/Network/BlockSynchronizer.cs
sgladkov: ok

src/Lachain.Core/Network/BlockSynchronizer.cs
tbssajal: Okay, but we sending request for 10 blocks ( max ). Is it okay? Or do we need to change it?
Depending on the txes in a block, execution time may vary. 4 secs could not be enough sometimes. Do we need to think about it for now?

sgladkov: ok, let use this vaule for now,  but add the task about measuring this time and adjusting it in the future

sgladkov (approved):

src/Lachain.Core/Network/BlockSynchronizer.cs
tbssajal: Ok

PR #297, opened 2022-10-20
Keygen spam check

tbssajal:
not tested yet

tested

### DIFF
--- a/src/Lachain.Consensus/ThresholdKeygen/Data/State.cs
+++ b/src/Lachain.Consensus/ThresholdKeygen/Data/State.cs
@@ -12,11 +12,13 @@ namespace Lachain.Consensus.ThresholdKeygen.Data
         public Commitment? Commitment { get; set; }
         public readonly Fr[] Values;
         public readonly bool[] Acks;
+        public bool Confirmation;
 
         public State(int n)
         {
             Values = new Fr[n];
             Acks = new bool[n];
+            Confirmation = false;
         }
 
         public int ValueCount()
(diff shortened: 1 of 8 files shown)


src/Lachain.Consensus/ThresholdKeygen/Data/State.cs
```diff
@@ -66,15 +69,17 @@ public static State FromBytes(ReadOnlyMemory<byte> bytes)
                 .Batch(Fr.ByteSize)
                 .Select(x => Fr.FromBytes(x.ToArray()))
                 .ToArray();
-            var acks = bytes.Slice(8 + cLen + n * Fr.ByteSize).ToArray()
+            var acks = bytes.Slice(8 + cLen + n * Fr.ByteSize, n).ToArray()
                 .Select(b => b != 0)
                 .ToArray();
+            var confirmation = bytes.Slice(8 + cLen + n * Fr.ByteSize + n).ToArray();
```
sgladkov, line 75: Have you tested new node synchronization from the genesis block after hardfork? I mean setup like this: start a network with old nodes,  than update them at some point to the new nodes,  generate some anount of blocks,  then delete blockchain data for one of the new blocks and try to sync it with other nodes.

tbssajal: Tested.

sgladkov (dismissed):

tbssajal:
Hardfork height must be set before keygen starts and after a cycle is finished, ie `hardfork_height % cycle_duration > 0` and `hardfork_height % cycle_duration < cycle_duration / 2`
Release has to be made following the same condition. When released: `current_height % cycle_duration > 0` and `current_height % cycle_duration < cycle_duration / 2`

soloway2010 (approved):

tbssajal:
needs hardfork

PR #298, opened 2022-10-21
Vault integration for wallet password

Anachor:
- Added new fields to `vaultConfig` 
  - `UseVault`: determines if vault will be used to store wallet password
  - `VaultAddress`: Address of the vault. Only used if `UseVault` is true. Will attempt to read `VAULT_ADDR` environment variable if `null` and `UseVault` is true.
  - `VaultToken`: Token to access the vault. Only used if `UseVault` is true. Will attempt to read `VAULT_TOKEN` environment variable if `null` and `UseVault` is true.
  - `VaultMountpoint`: Mountpoint` of vault. Only used if `UseVault` is true.
  - `VaultEndpoint`: Path of vault where password will be stored. Only used if `UseVault` is true.
  
- If `UseVault` is `true` is config, will attempt to read password from vault using above parameters.
- Otherwise, will attempt to read from `password` field in config. 
- Implemented `Change Password` for both cases.

### DIFF
--- a/src/Lachain.Console/TrustedKeygen.cs
+++ b/src/Lachain.Console/TrustedKeygen.cs
@@ -157,6 +157,7 @@ namespace Lachain.Console
                 var walletPath = "wallet.json";
                 var vault = new VaultConfig
                 {
+                    UseVault = false,
                     Path = walletPath,
                     Password = getRandomPassword()
                 };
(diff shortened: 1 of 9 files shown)


sgladkov (approved):
LGTM, didn't tested

PR #299, opened 2022-10-21
fixed bug of waiting for non existing tx

tbssajal:
a small bug: even if tx is not added to pool, validator status manager waits for its confirmation


sgladkov:
So,  we have cases when we fail to add system tx to pool?

tbssajal:
> So, we have cases when we fail to add system tx to pool?

I did not find any tx that failed to add to pool, but the sender is the node, so it could be failed due to nonce error or something else, maybe insufficient balance.

sgladkov (approved):

PR #303, opened 2022-11-02
Consensus persistence

Anachor:


src/Lachain.Core/DI/Modules/ConsensusModule.cs
```diff
@@ -16,6 +17,7 @@ public void Register(IContainerBuilder containerBuilder, IConfigManager configMa
             containerBuilder.RegisterSingleton<IValidatorManager, ValidatorManager>();
             containerBuilder.RegisterSingleton<IKeyGenManager, KeyGenManager>();
             containerBuilder.RegisterSingleton<IValidatorStatusManager, ValidatorStatusManager>();
+            containerBuilder.Register<IMessageRepository, RootProtocolMessageRepository>();
```
tbssajal, line 20: maybe this should be registered in storage module

tbssajal (changes requested):

PR #307, opened 2022-11-07
Fix returm format for eth_getStorageAt API call

sgladkov:


sgladkov:
it is wrong solution,  previous format was correct

PR #311, opened 2022-11-14
Consensus message request

tbssajal:


Anachor (changes requested):
There is a Protocol Type class implemented for Consensus Persistence as well. If both prs are merged this will be a duplication and may cause errors. It should be made consistent.

tbssajal:
> There is a Protocol Type class implemented for Consensus Persistence as well. If both prs are merged this will be a duplication and may cause errors. It should be made consi

PR #318, opened 2022-11-25
Fix keygen phase detection

tbssajal:
tested


tbssajal:
As more than 2/3 of validators participate to produce blocks, median height should give the correct height if malicious nodes give wrong heights.

soloway2010 (approved):

PR #338, opened 2023-01-25
Fix contract balance transfer

tbssajal:
tested with hardfork


tbssajal:
Need to calculated hardfork 16 height before release

soloway2010 (dismissed):

